PR #5, opened 2019-07-30
Add Trusted Compute Framework Code

danintel:
Signed-off-by: daniela1 <daniel.anderson@intel.com>


danintel:
Ran sanity check.
Also diffed with internal repository--no differences.

PR #6, opened 2019-07-30
Add Tests for Passing "-" as encryptedDataEncryptionKey

danintel:
* Add tests json_{10,11,12,13}.json
* Replace invalid hex string "mhash555" with a valid example

Signed-off-by: acmanjun <manjunath.a.c@intel.com>


danintel:
The json_1*.json files were missed in the original commit, and fixing the hex string.

PR #10, opened 2019-08-06
Refactor enclave, manager and utilities

manju956:
Includes fixes for following
- Rename work_load_id to workload_id
- Fix incorrect json error message when encryptedRequestHash field is missing in request
- Add return values for functions decrypt_data and decrypted_response
- Provide proper hex string values for ApplicationTypeId and OrganizationId
- worker_id is made valid hex string


docker-compose.yaml
```diff
@@ -16,7 +16,7 @@ version: '3.5'
 
 services:
   tcf:
-    container_name: tcf
+    container_name: tcf-manju
```
danintel: I suggest restoring the name `tcf` or using another name than` tcf-manju`. Was this for debugging?

config/tcs_config.toml
```diff
@@ -66,8 +66,8 @@ WorkOrderSyncUri = "http://localhost:8080"
 WorkOrderAsyncUri = "http://localhost:8080"
 WorkOrderPullUri = "http://localhost:8080"
 WorkOrderNotifyUri = "http://localhost:8080"
-ApplicationTypeId = "Application_01"
```
Ram-srini: Instead adding hard coded random value, can you add something like in the python code.
"Application_01".encode("UTF-8").hex()
"Organization_01".encode("UTF-8").hex()

docker-compose.yaml
manju956: this file was mistakenly added in the commit list. I will remove it.

config/tcs_config.toml
manju956: If enclave manager does the hex conversion, aren't we modifying client specific config?
As per spec, users need to input their applicationtypeId and OrganizationId in hex format. I will add this comment.

PR #11, opened 2019-08-06
Utility to create error response is added

manju956:
- A common utility function which accepts error code and message to create
  JRPC error response object is added. This eliminates redundant code.

### DIFF
--- a/tcs/tcs_listener/tcs_listener.py
+++ b/tcs/tcs_listener/tcs_listener.py
@@ -41,6 +41,7 @@ from tcs_workorder_receipt_handler import TCSWorkOrderReceiptHandler
 from tcs_worker_encryption_key_handler import WorkerEncryptionKeyHandler
 from shared_kv.shared_kv_interface import KvStorage
 from error_code.error_status import WorkorderError
+import utils.utility as utility
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


tcs/tcs_listener/tcs_listener.py
```diff
@@ -119,21 +120,17 @@ def _process_request(self, input_json_str):
             return response
 
     def render_GET(self, request):
-        response = {}
-        response['error'] = {}
-        response['error']['code'] = WorkorderError.INVALID_PARAMETER_FORMAT_OR_VALUE
-        response['error']['message'] = 'Only POST request is supported'
+        response = utility.create_error_response(WorkorderError.INVALID_PARAMETER_FORMAT_OR_VALUE, "0", \
```
danintel: Just so you know, in Python, a "\" at the end of the line is not needed to continue function parameters on the next line.
 Python knows there are more parameters because the "(" is not closed yet with ")". In other words, you can leave out the "\" at the end of the line (here and elsewhere).

manju956: thanks for the info, i will remove it.

tcs/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -303,11 +310,12 @@ def __process_worker_update(self, worker_id, input_json_str, response):
...
-            response['error']['message'] = "Successfully Updated"
+            response = utility.create_error_response(
+                    WorkerError.SUCCESS, jrpc_id,
+                    "Successfully Updated")
         else :
-
-            response['error']['code'] = WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE
-            response['error']['message'] = "Worker Id not found in the database. Hence invalid parameter"
+            response = utility.create_error_response(
+                    WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE, json_id,
```
manojgop: json_id -> jrpc_id ?

tcs/tcs_listener/tcs_listener.py
```diff
@@ -119,21 +120,17 @@ def _process_request(self, input_json_str):
...
 
     def render_POST(self, request):
         response = {}
-        response['error'] = {}
-        response['error']['code'] = WorkorderError.UNKNOWN_ERROR
-       
-        logger.info('Received a new request from the client')
 
+        logger.info('Received a new request from the client')
+        jrpc_id = "0"
```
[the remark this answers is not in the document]
manju956: In case of get request, we don't have request body which contains "id". hence jrpc_id is hard coded to 0.

tcs/tcs_listener/tcs_worker_registry_handler.py
manju956: Done

PR #15, opened 2019-08-07
Use workload_id param to identify type of workorder requests

manju956:
Includes following changes
- Remove index 0 from all input requests
- Use workload_id as identifier of work order requests


tcs/core/tcs_trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -213,31 +213,28 @@ namespace tcf {
...
+                in_wo_data.emplace_back(d.workorder_data.index, d.workorder_data.decrypted_data);
+            }
+
+            for (auto d : data_items_out) {
+                out_wo_data.emplace_back(d.workorder_data.index, d.workorder_data.decrypted_data);
+            }
+
+            WorkloadProcessor processor;
+            std::string workload_str;
+            HexStringToBinary((uint8_t *)workload_str.c_str(), work_load_id.length(), work_load_id);
```
manojgop: Where is memory allocation done for first argument of HexStringToBinary

manju956: I will correct it.

PR #24, opened 2019-08-16
Make dataHash parameter as empty value

manju956:
- Since dataHash parameter is optional, keep it as empty value.

### DIFF
--- a/tests/json_requests/json_03.json
+++ b/tests/json_requests/json_03.json
@@ -11,7 +11,7 @@
         "workerId":"",
         "workloadId": "heart-disease-eval",
         "requesterId": "0x3456",
-        "workerEncryptionKey": "0x6789",
+        "workerEncryptionKey": "",
         "dataEncryptionAlgorithm": "AES-GCM-256",
         "encryptedSessionKey": "sessionkey",
         "sessionKeyIv": "Iv",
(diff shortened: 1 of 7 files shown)


tests/work_orders/04_work_order_submit.json
```diff
@@ -20,21 +20,15 @@
...
         "inData": [
                    {"index": 1,
-                    "dataHash": "mhash444",
+                    "dataHash": "",
                     "data": "Heart disease evaluation data: 32 1 1 156  132 125 1 95  1 0 1 1 3 1",
                     "encryptedDataEncryptionKey": "",
                     "iv": ""
                    }
         ],
         "outData": [
```
Ram-srini: Do we need to pass the outData in work order submit request? I don't think it is required. Can you remove this as well.?

tests/json_requests/json_03.json
```diff
@@ -20,21 +20,21 @@
         "requesterSignature": "",
         "inData": [
 		   {"index": 2,
-                    "dataHash": "12ab34cd",
+                    "dataHash": "",
```
manojgop: workerEncryptionKey is also hardcoded in json file. This could also be removed. Either do in in same PR and update the commit message or raise a new PR

manju956: Ok.

tests/work_orders/04_work_order_submit.json
manju956: Ok

PR #25, opened 2019-08-16
Move workload related classes from dispatcher to respective workloads

manju956:
- Move EchoResult and HeartDiseaseEvalFactory classes from
  dispatcher to respective workloads. This decouples dispacther code from workloads

### DIFF
--- a/common/sgx_workload/workload_dispatcher.cpp
+++ b/common/sgx_workload/workload_dispatcher.cpp
@@ -19,97 +19,12 @@
 #include "echo/workload/echo.h"
 #include "heart_disease_eval/workload/heart_disease_evaluation.h"
 
-ByteArray ConvertStringToByteArray(std::string s) {
-    ByteArray ba(s.begin(), s.end());
-    return ba;
-}
-
-class EchoResult: public tcf::WorkOrderProcessorInterface {
-public:
-        void ProcessWorkOrder(
-                std::string workload_id,
-                const ByteArray& participant_address,
-                const ByteArray& enclave_id,
-                const ByteArray& work_order_id,
-                const std::vector<tcf::WorkOrderData>& in_work_order_data,
-                std::vector<tcf::WorkOrderData>& out_work_order_data) {
-        std::string result_str;
-        int i = 0;
-        int out_wo_data_size = out_work_order_data.size();
-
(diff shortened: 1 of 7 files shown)


examples/apps/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 3.2 FATAL_ERROR)
 
 INCLUDE(CMakeVariables.txt)
+SET(GENERIC_PRIVATE_INCLUDE_DIRS "." "${TCF_TOP_DIR}/tc/sgx/common"
+	"${TCF_TOP_DIR}/tc/sgx/common/work_order_interface")
```
manojgop: What's the dependency with ${TCF_TOP_DIR}/tc/sgx/common"

examples/apps/echo/workload/echo_factory.cpp
```diff
@@ -0,0 +1,58 @@
+
```
manojgop: No need of echo_factor.h and echo_factory.cpp. These could be moved to echo.h. EchoResultImpl  in echo.h is not required. Process in EchoResultImpl  can be a private function in EchoResult. 
Similarly no need of heart_disease_evaluation_factory.cpp

common/sgx_workload/workload_dispatcher.cpp
```diff
@@ -16,96 +16,18 @@
 #include <string>
 #include "work_order_data.h"
 #include "workload_processor.h"
-#include "echo/workload/echo.h"
-#include "heart_disease_eval/workload/heart_disease_evaluation.h"
+#include "echo/workload/echo_factory.h"
+#include "heart_disease_eval/workload/heart_disease_evaluation_factory.h"
```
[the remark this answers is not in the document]
manju956: Ok

examples/apps/CMakeLists.txt
manju956: work_order_interface includes error.h from ${TCF_TOP_DIR}/tc/sgx/common.

examples/apps/echo/workload/echo_factory.cpp
manju956: Ok

PR #28, opened 2019-08-20
Fix worker registry crash

manju956:
When worker update flow is called, worker registry handler crashes.
Removed wrong JSON deserialization


examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -312,7 +312,6 @@ def __process_worker_update(self, worker_id, input_json_str, response):
             json_dict = json.loads(value)
             input_value = json.loads(input_json_str)
             worker_details = input_value ["params"]["details"]
-            worker_details = json.loads(worker_details)
```
Ram-srini: Worker details field would be string and converting it to json object. Worker registry update will update json object and convert it back to string and updating details field at 318. Which test is failing?

manju956: I believe details param is dictionary not string. that's why json.loads is failing.

PR #40, opened 2019-08-27
fix contract wrapper and readme

m3ngyang:
I ran the ethereum worker registry contract unit tests and found some problems:
- current codes do not comply with the deployed contracts, which causes the failure of transactions. Users may need to deploy brand new contract themself, 
- there is a bug in the `ethereum_wrapper.py`, namely representing tx hash without `.hex()` transformation.

I fixed the mentioned bugs and also embellished the readme file.

### DIFF
--- a/examples/common/python/connectors/TestingContracts.md
+++ b/examples/common/python/connectors/TestingContracts.md
@@ -1,27 +1,60 @@
-Requirements:
+# Test Process
+
 1. Installing Python 3.6.8
+
+    ```bash
     wget https://www.python.org/ftp/python/3.6.8/Python-3.6.8.tgz
     tar -xvf Python-3.6.8.tgz
     cd Python-3.6.8
     ./configure
-    make 
+    make
     sudo make install
-    make sure 
+    make sure
+    ```
+
 2. Install web3.py an Ethereum python client interact with Ethereum network(About web3.py - https://web3py.readthedocs.io/en/stable/quickstart.html)
+
+    ```bash
     sudo pip install web3
+    ```
+
 3. Install solc to compile solidity contract from python
+
+    ```bash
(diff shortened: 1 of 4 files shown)


m3ngyang:
fixup issue #41

examples/common/python/connectors/tcf_connector.toml
```diff
@@ -33,12 +33,12 @@ direct_registry_contract_file = "examples/common/python/connectors/ethereum/cont
 worker_registry_contract_file = "examples/common/python/connectors/ethereum/contracts/WorkerRegistry.sol"
 #Deployed contract address of direct registry contract address.
 #Initial deploy the contract using eth_cli.py to get the this address
-direct_registry_contract_address = "0x9Be28B132aeE1b2c5A1C50529a636cEd807842cd"
+direct_registry_contract_address = ""
```
Ram-srini: If we set the direct_registry_contract_address to empty our unit tests won't run. That's why we added common shared account for testing contracts.

m3ngyang: That makes sense and addresses of runnable pre-installed contracts can be reserved.

Ram-srini: It shouldn't fail we are using those contract instances for our testing. It is shared account, if you want test with specific account then update with your credentials and contract instances locally.

m3ngyang: According to the given ethereum contracts, most functions must be called by the contract deployer. If the these default address are reserved, other users must deploy their own contracts and replace the related addresses in the config file.

m3ngyang:
@Ram-srini default addresses are reserved in the new patch.

PR #34, opened 2019-08-22
add an optional KvStorage in form of Remote LMDB [PR#21] module for TCSListener and EnclaveManager

sammyne:
1. bootstrap the LMDB as an independent service in scripts/tcs_startup.sh if `LMDBHelperProxy` is employed
    2. add `LMDBHelperProxy` as an remote option to interact with LMDB for TCSListener and EnclaveManager

Signed-off-by: Sammy <lixiangmin01@baidu.com>

PR #54, opened 2019-09-04
unify the API for connecting to local/remote LMDB

sammyne:


danintel:
The remote LMDB was added with PR #21 and PR #34

PR #58, opened 2019-09-06
Add Documentation for the Echo Demo Application

danintel:
* Add file examples/apps/echo/README.md that documents how to use
  the echo demo application
* Make script examples/.../echo_client.py executible

### DIFF
--- a/examples/apps/echo/README.md
+++ b/examples/apps/echo/README.md
@@ -1,0 +1,49 @@
+# Echo Demo Application
+
+This simple demo application echos back the same message sent from
+the client to the worker.
+
+## Using the Echo Command Line Client
+
+The command line client, `echo_client.py` sends a message to the worker,
+which echos back the same message back to the client.
+For command line options, type `./echo_client.py -h` from the
+`client` directory.
+
+To use:
+
+1.  If needed, update the Ethereum account and direct registry contract
+    information in `docker/Dockerfile.tcf-dev` and
+    `examples/common/python/connectors/tcf_connector.toml`
+2.  Follow instructions in the "Docker-based Build and Execution" section of
+    the [build document](../../../BUILD.md#dockerbuild) through step 4
+    (activating a virtual environment)
+3.  Terminal 1 is running the TCF Enclave Manager and Listener with
+    `docker-compose` . Terminal 2 is running the Docker container shell
+    with the `(_dev)` prompt
+4.  In Terminal 2 run `cd $TCF_HOME/examples/apps/echo/client`
+5.  In Terminal 2 install the Solidity compiler as follows:
+    ```bash
+    mkdir -p $HOME/.py-solc/solc-v0.4.25/bin \
+    && curl -LsS https://github.com/ethereum/solidity/releases/download/v0.4.25/solc-static-linux \
+            -o /root/.py-solc/solc-v0.4.25/bin/solc \
+    && chmod 0755 /root/.py-solc/solc-v0.4.25/bin/solc &&
+    export SOLC_BINARY=/root/.py-solc/solc-v0.4.25/bin/solc
+    ```
+6. In Terminal 2, set environment variable `WALLET_PRIVATE_KEY` if not set.
+    This should match the value in file `docker/Dockerfile.tcf-dev`
+    from step 3 above:
+    ```bash
+    export WALLET_PRIVATE_KEY="B413189C95B48737AE2D9AF4CAE97EB03F4DE40599DF8E6C89DCE4C2E2CBA8DE"
+    ```
+7.  In Terminal 2, run `./echo_client.py -m "Hello world"` .
+    Use the `-h` option to see other available options
+
+8.  You will see output showing:
+    1. The client searches the registry for an "echo" worker
+    2. Sends a request to the worker
+    3. Waits for and receives a response.
+    4. The messages sent and received are encrypted with the key specified
+       in the JSON packet, then encoded in Base64.
+9.  In Terminal 1, press Ctrl-c to stop the TCF Enclave Manager and Listener
+
--- a/examples/apps/echo/README.md
+++ b/examples/apps/echo/README.md
@@ -22,28 +22,14 @@ To use:
     `docker-compose` . Terminal 2 is running the Docker container shell
     with the `(_dev)` prompt
 4.  In Terminal 2 run `cd $TCF_HOME/examples/apps/echo/client`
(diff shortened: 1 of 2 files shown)


examples/apps/echo/README.md
```diff
@@ -0,0 +1,53 @@
...
+2.  Follow instructions in the "Docker-based Build and Execution" section of
+    the [build document](../../../BUILD.md#dockerbuild) through step 4
+    (activating a virtual environment)
+3.  Terminal 1 is running the TCF Enclave Manager and Listener with
+    `docker-compose` . Terminal 2 is running the Docker container shell
+    with the `(_dev)` prompt
+4.  In Terminal 2 run `cd $TCF_HOME/examples/apps/echo/client`
+5.  In Terminal 2 install Python3's TKInter GUI library with
+    ```bash
+    export DEBIAN_FRONTEND=noninteractive; apt update; apt -y install python3-tk
```
manojgop: Is echo client using GUI library ? AFAIK only Heart Disease eval client is using GUI.
python3-tk  is already installed in docker. Is this instruction for standalone build ?

danintel: @manojgop You are correct. The echo client is not using the TK GUI library, so I removed #5 above. Thanks.

examples/apps/echo/README.md
```diff
@@ -0,0 +1,49 @@
...
+    information in `docker/Dockerfile.tcf-dev` and
+    `examples/common/python/connectors/tcf_connector.toml`
+2.  Follow instructions in the "Docker-based Build and Execution" section of
+    the [build document](../../../BUILD.md#dockerbuild) through step 4
+    (activating a virtual environment)
+3.  Terminal 1 is running the TCF Enclave Manager and Listener with
+    `docker-compose` . Terminal 2 is running the Docker container shell
+    with the `(_dev)` prompt
+4.  In Terminal 2 run `cd $TCF_HOME/examples/apps/echo/client`
+5.  In Terminal 2 install the Solidity compiler as follows:
```
manojgop: Solidity compiler is already installed in dockerfile in docker/Dockerfile.tcf-dev

examples/apps/echo/README.md
```diff
@@ -0,0 +1,49 @@
...
+    ```bash
+    mkdir -p $HOME/.py-solc/solc-v0.4.25/bin \
+    && curl -LsS https://github.com/ethereum/solidity/releases/download/v0.4.25/solc-static-linux \
+            -o /root/.py-solc/solc-v0.4.25/bin/solc \
+    && chmod 0755 /root/.py-solc/solc-v0.4.25/bin/solc &&
+    export SOLC_BINARY=/root/.py-solc/solc-v0.4.25/bin/solc
```
manju956: Can we add these env variables to /etc/environment so that these variables will be auto set for every docker shells?

examples/apps/echo/README.md
```diff
@@ -0,0 +1,49 @@
...
+    ```
+6. In Terminal 2, set environment variable `WALLET_PRIVATE_KEY` if not set.
+    This should match the value in file `docker/Dockerfile.tcf-dev`
+    from step 3 above:
+    ```bash
+    export WALLET_PRIVATE_KEY="B413189C95B48737AE2D9AF4CAE97EB03F4DE40599DF8E6C89DCE4C2E2CBA8DE"
```
manju956: Can we add these env variables to /etc/environment so that these variables will be auto set for every docker shells?

examples/apps/echo/README.md
danintel: No it is not. It is not there for the bash shell and it has to be installed manually.
For example:

```
$ sudo docker exec -it tcf bash
root@f48d605bea9e:/project/TrustedComputeFramework/tools/build#  source _dev/bin/activate
(_dev) root@f48d605bea9e:/project/TrustedComputeFramework/tools/build# ls -a ~/.py-solc
ls: cannot access '/root/.py-solc': No such file or directory
(_dev) root@f48d605bea9e:/project/TrustedComputeFramework/tools/build# echo $SOLC_BINARY
```

examples/apps/echo/README.md
danintel: The problem is the solc compiler is not installed for the docker bash shell (see above).

If that problem is fixed, how do I add it?  Do I edit `/etc/environment` somewhere?
Or do I just add a line in `docker-compose.yaml` under the `environment:` stanza?

examples/apps/echo/README.md
danintel: Yes.  See above.

examples/apps/echo/README.md
manojgop: So although solc is installed from dockerfile file, it's not accessible from other shell. May be we coudl eitehr insatll solc in /usr/bin or add the solc path of /etc/environment  or ~/.bashrc so that it's accessible from all bash shell. @Ram-srini Can this be done ?

Ram-srini: @danintel for docker based build we don't need to install solc compiler again.
(_dev) root@0840863d8f74:/project/TrustedComputeFramework/tools/build# pip list web3 |grep web3
web3                5.0.2
(_dev) root@0840863d8f74:/project/TrustedComputeFramework/tools/build# echo $SOLC_BINARY
/root/.py-solc/solc-v0.4.25/bin/solc
(_dev) root@0840863d8f74:/project/TrustedComputeFramework/tools/build# ls $SOLC_BINARY
/root/.py-solc/solc-v0.4.25/bin/solc

examples/apps/echo/README.md
Ram-srini: I am able to run echo client with Ethereum with below option
_dev) root@0840863d8f74:/project/TrustedComputeFramework/examples/apps/echo/client# python echo_client.py --config ../../../common/python/connectors/tcf_connector.toml

examples/apps/echo/README.md
danintel: I experimented around on different machines.  I found this stuff works only on a machine with direct Internet access. It does not work behind the Intel corporate proxy.  I created this JIRA bug for this problem:

danintel: I found the problem. The `BUILD.md` instructions are missing:
`docker-compose up --build`
Otherwise the `docker/Dockerfile.tcf-dev` file is not ran.
This occurs if an old Docker image still exists.
I am fixing steps in `docker/Dockerfile.tcf-dev` as the OpenSSL build fails.  I think the problem is `$THREADS` is causing a failure.

PR #59, opened 2019-09-09
feature(make): add a Makefile to speed up the local development

sammyne:
Add a Makefile for the local development to **avoid rebuilding the redundant virtual environment every time**

### DIFF
--- a/tools/build/Makefile.dev
+++ b/tools/build/Makefile.dev
@@ -1,0 +1,15 @@
+# This serves for the local development only. After the first building by 
+# `make`, running `make -f Makefile.dev` should avoid the unnecessary 
+# rebuilding of the python virtual environment every time.
+
+SRCDIR?=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+DSTDIR?=$(abspath $(SRCDIR)/_dev)
+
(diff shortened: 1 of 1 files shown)


tools/build/Makefile.dev
```diff
@@ -0,0 +1,15 @@
+# This serves for the local development only. After the first building by
```
manojgop: Will this help with docker build or only standalone build ?

sammyne: I don't understand what the standalone build is ~

The intention is to help such a scenario that updates codes and then rebuilds for testing locally. In such scenario, the initial setup (including upgrading tools such `pip`) for `virtualenv` should work all along and requires no redo every time, and we just need to run the `${TCF_HOME}/scripts/rebuild.sh` to rebuild the whole project

manojgop: By standalone build, I mean build performed outside docker (i.e not via docker-compose up). I'm thinking would it make more sense to move the pip install packages in tools/build/Makefile to Dockerfile and make it part of docker image (i.e use apt-install of python pip packages)

sammyne: Gotcha~ 

I found that running `make packages` work equivalently with this PR, so it's closed.

Moving the virtualenv setup into Dockerfile is a good . I'll work on that later.

PR #64, opened 2019-09-11
simplify the routine for looking up workers

sammyne:

### DIFF
--- a/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
+++ b/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
@@ -14,7 +14,6 @@
 
 import json
 import logging
-from itertools import cycle
 import utility.utility as utility
 from error_code.error_status import WorkerError
 from error_code.error_status import WorkerStatus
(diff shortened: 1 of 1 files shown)


examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -184,62 +183,47 @@ def __process_worker_set_status(self, worker_id, input_json_str, response):
...
         for worker_id in self.worker_pool:
-            if lookup_bool:
-                if (worker_id == input_value["params"]["lookUpTag"]):
-                    lookup_bool = False
-                    continue
-                else:
-                    continue
+            if is_lookup_next:
+                # loop until found the expected lookUpTag
+                is_lookup_next = (worker_id != params["lookUpTag"])
```
danintel: On line 199 you use `lookUpTag`:
> ` 199 is_lookup_next = (worker_id != params["lookUpTag"])`
but on line 223 you use "`lookupTag`
> `223 "lookupTag": lookupTag,`
These should be the same.  Looking at other source files, `lookUpTag` is only used here, so change `lookUpTag` to `lookupTag`

sammyne: done

Issue #51, opened 2019-09-03
lack module EthereumDirectRegistry in direct_registry_bridge.py


m3ngyang:
`direct_registry_bridge.py` doesn't work because of lacking import module `EthereumDirectRegistry`.

m3ngyang:
I'm doing some tests on the bridge module, but it seems that some files are missing, can you provide them? @danintel @manju956

danintel:
`EthereumDirectRegistry` is used by:
`examples/common/python/connectors/ethereum/unit_tests/test_direct_registry_bridge.py`
`examples/common/python/connectors/ethereum/direct_registry_bridge.py`
It is implemented with the Solidity source file
`./examples/common/python/connec
tors/ethereum/contracts/WorkerRegistryList.sol`

For details, see https://github.com/hyperledger-labs/trusted-compute-framework/blob/master/examples/common/python/connectors/TestingContracts.md

PR #65, opened 2019-09-12
[WIP] fix up bridge module bugs

m3ngyang:
bridge module doesn't work currently. This pr aims to fix the problem. #51

### DIFF
--- a/config/tcs_config.toml
+++ b/config/tcs_config.toml
@@ -14,7 +14,7 @@
 
 [KvStorage]
 StoragePath = "config/Kv_Shared_tmp"
-StorageSize = "1 TB"
+StorageSize = "1 GB"
 # the remote version is of higher priority if enabled
 #remote_url = "http://localhost:9090"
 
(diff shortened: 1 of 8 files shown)


config/tcs_config.toml
```diff
@@ -14,7 +14,7 @@
 
 [KvStorage]
 StoragePath = "config/Kv_Shared_tmp"
-StorageSize = "1 TB"
+StorageSize = "1 GB"
```
danintel: 1 GByte is very small. The space for the file is not allocated until used, so 1 TByte allows room to grow.

danintel (changes requested):

config/tcs_config.toml
m3ngyang: 1 GB is only for local test, which won't be included in the final commit.

Ram-srini:
@m3ngyang Can you please update your commit with sign off message.

m3ngyang:
@Ram-srini singed off. but this pr is not finished.

PR #68, opened 2019-09-12
Fix docker/Dockerfile.tcf-dev File

danintel:
* Do not recursively `source /etc/environment` in itself
* Add `ENV SOLC_BINARY` to `/etc/environment`
* Add `WALLET_PRIVATE_KEY` to `/etc/environment`
* Move py-solc build out of the middle of the OpenSSL build
* Remove tar -v flag
* Minor formatting fixes

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -39,7 +39,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # Add necessary packages
 RUN apt-get update \
- && apt-get install -y -q\
+ && apt-get install -y -q \
     build-essential \
     ca-certificates \
     cmake \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.tcf-dev
```diff
@@ -86,33 +86,34 @@ RUN wget https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64
...
  && cd .. \
- && rm -rf openssl-1.1.1d
-# Note: we do _not_ delete openssl-1.1.1d.tar.gz as we re-use it below ..
-
-# py-solc 3.2.0 requires an older version of the solc compiler, 0.4.25
-RUN mkdir -p $HOME/.py-solc/solc-v0.4.25/bin \
-  && curl -LsS https://github.com/ethereum/solidity/releases/download/v0.4.25/solc-static-linux -o /root/.py-solc/solc-v0.4.25/bin/solc \
-  && chmod 0755 /root/.py-solc/solc-v0.4.25/bin/solc
+ && rm -rf openssl-$OPENSSL_VER
+# Note: we do _not_ delete openssl-*.tar.gz as we re-use it below ..
```
manojgop: Can't we use openssl-$OPENSSL_VER.tar.gz

docker/Dockerfile.tcf-dev
```diff
@@ -86,33 +86,34 @@ RUN wget https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64
...
 
 # ("trusted") SGX OpenSSL
 # Note: This will compile in HW or SIM mode depending on the availability of
 # /dev/isgx and /var/run/aesmd/aesm.socket
 WORKDIR /tmp
 RUN git clone https://github.com/intel/intel-sgx-ssl.git  \
  && . /opt/intel/sgxsdk/environment \
- && (cd intel-sgx-ssl/openssl_source; mv /tmp/openssl-1.1.1d.tar.gz . ) \
+ && (cd intel-sgx-ssl/openssl_source; mv /tmp/openssl-*.tar.gz . ) \
```
manojgop: can we use  `mv openssl-$OPENSSL_VER.tar.gz .`

docker/Dockerfile.tcf-dev
danintel: No. This is removing the build directory, not the tar file. So `openssl-$OPENSSL_VER` (without `.tar.gz`) is correct. The tar file is needed below as mentioned in the comment on line 108:
`# Note: we do _not_ delete openssl-*.tar.gz as we re-use it below ..`

docker/Dockerfile.tcf-dev
danintel: We can. I have to define OPENSSL_VER again because the previous value was lost with the previous `RUN` command.

docker/Dockerfile.tcf-dev
manojgop: ok. instead of * can we use $OPENSSL_VER

delete openssl-*.tar.gz -> delete openssl-$OPENSSL_VER.tar.gz

danintel: Done. I had to combine the `RUN` statements so `$OPENSSL_VER` is still defined.

PR #76, opened 2019-09-17
Create Initial docs Directory

danintel:


danintel:
I will add a tutorial here later, after this is PR pushed.

PR #83, opened 2019-09-22
Add Validation and handle exceptions

manju956:
- Handle exceptions in generate_signaure method
- Add validations for dataEncryptionAlgorithm, Payload format,
  empty inData in work_order_processor
- Use requesterNonce if passed in the input request
- Add checks for validating worker status in the input request

### DIFF
--- a/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
+++ b/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
@@ -130,7 +130,6 @@ class TCSWorkerRegistryHandler:
         if(self.kv_helper.get("workers", worker_id) is None):
             input_value = {}
             input_value = input_value_json['params']
-
             # Worker Initial Status is set to Active
             input_value["status"] = WorkerStatus.ACTIVE
 
@@ -146,8 +145,23 @@ class TCSWorkerRegistryHandler:
                 "Worker Id already exists in the database. \
                  Hence invalid parameter")
         return response
-# ------------------------------------------------------------------------------------------------
 
+# ------------------------------------------------------------------------------------------------
+    def validate_input_worker_status(self, input_json):
+        '''
+        Validates worker status in the input request
+        '''
+        try:
+            worker_status = input_json["params"]["status"]
+            status = int(worker_status)
+            if status < WorkerStatus.ACTIVE.value or status > WorkerStatus.COMPROMISED.value:
+                return WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE
+        except:
+            logger.error("Invalid parameter format or value found in the request")
+            return WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE
+        return WorkerError.SUCCESS
+
+# ------------------------------------------------------------------------------------------------
     def __process_worker_set_status(self, worker_id, input_value, response):
         """
         Function to set the status of worker
@@ -164,8 +178,14 @@ class TCSWorkerRegistryHandler:
         value = self.kv_helper.get("workers", worker_id)
         if value:
             json_dict = json.loads(value)
-            json_dict['status'] = input_value['params']['status']
+            if self.validate_input_worker_status(input_value) !=  WorkerError.SUCCESS:
+                response = utility.create_error_response(
+                           WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE,
+                           jrpc_id,
+                           "Invalid parameter format or value found in the request")
+                return response
 
+            json_dict['status'] = input_value['params']['status']
             value = json.dumps(json_dict)
             self.kv_helper.set("workers", worker_id, value)
             response = utility.create_error_response(
--- a/examples/common/python/utility/signature.py
+++ b/examples/common/python/utility/signature.py
@@ -177,16 +177,20 @@ class ClientSignature(object) :
     def generate_signature(self, hash, private_key):
         """
         Function to generate signature object
+        Return a tuple containing signature and status
         Parameters:
             - hash is the combined array of all hashes calculated on the message
             - private_key is Client private key
         """
-
-        self.private_key = private_key
-        self.public_key =  self.private_key.GetPublicKey().Serialize()
-        signature_result =  self.private_key.SignMessage(hash)
-        signature_base64  =  crypto.byte_array_to_base64(signature_result)
-        return  signature_base64
+        try:
+            self.private_key = private_key
+            self.public_key =  self.private_key.GetPublicKey().Serialize()
+            signature_result =  self.private_key.SignMessage(hash)
+            signature_base64  =  crypto.byte_array_to_base64(signature_result)
+        except:
+            logger.error("Exception occured during signature generation")
+            return (None, SignatureStatus.FAILED)
+        return  signature_base64, SignatureStatus.PASSED
 
 #---------------------------------------------------------------------------------------------
     def generate_client_signature(self, input_json_str,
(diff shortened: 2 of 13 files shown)


examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -146,8 +145,20 @@ def __process_worker_register(self, worker_id, input_value_json, response):
                 "Worker Id already exists in the database. \
                  Hence invalid parameter")
         return response
+
 # ------------------------------------------------------------------------------------------------
+    def validate_input_worker_status(self, input_json):
+        try:
+            worker_status = input_json["params"]["status"]
+            status = int(worker_status)
+            if status < 0 or status > 4:
```
manojgop: Can we use enum for status instead of "4"

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -146,8 +145,20 @@ def __process_worker_register(self, worker_id, input_value_json, response):
                 "Worker Id already exists in the database. \
                  Hence invalid parameter")
         return response
+
 # ------------------------------------------------------------------------------------------------
+    def validate_input_worker_status(self, input_json):
+        try:
```
manojgop: Please add function definition comments

examples/common/python/utility/signature.py
```diff
@@ -181,11 +181,14 @@ def generate_signature(self, hash, private_key):
...
-        signature_result =  self.private_key.SignMessage(hash)
-        signature_base64  =  crypto.byte_array_to_base64(signature_result)
+        try:
+            self.private_key = private_key
+            self.public_key =  self.private_key.GetPublicKey().Serialize()
+            signature_result =  self.private_key.SignMessage(hash)
+            signature_base64  =  crypto.byte_array_to_base64(signature_result)
+        except:
+            logger.error("Exception occured during signature generation")
+            return SignatureStatus.FAILED
```
manojgop: Better to return consistent types on success and failure. One option is to return a tuple with (signature, status). Also please update function comments accordingly

examples/common/python/utility/signature.py
```diff
@@ -230,9 +233,13 @@ def generate_client_signature(self, input_json_str,
         encrypted_session_key_str = ''.join(format(i, '02x') for i in encrypted_session_key)
         self.__encrypt_workorder_indata(input_json_params, session_key,
                 session_iv, worker.encryption_key, data_key, data_iv)
-        # [NO_OF_BYTES] 16 BYTES for nonce, is the recommendation by NIST to
-        # avoid collisions by the "Birthday Paradox".
-        nonce =  crypto.random_bit_string(NO_OF_BYTES)
+
+        if input_json_params["requesterNonce"]:
+            nonce = crypto.string_to_byte_array(input_json_params["requesterNonce"])
```
manojgop: Are we checking elsewhere if input_json_params["requesterNonce"] is of type hex string. will crypto.string_to_byte_array() throw exception if input is not string. Then its good to catch the exception

examples/common/python/utility/signature.py
```diff
@@ -256,7 +263,10 @@ def generate_client_signature(self, input_json_str,
 
         #Update the input json params
         input_json_params["encryptedRequestHash"] = encrypted_request_hash_str
-        input_json_params['requesterSignature'] = self.generate_signature(final_hash, private_key)
+        signature = self.generate_signature(final_hash, private_key)
+        if signature == SignatureStatus.FAILED:
+            return SignatureStatus.FAILED
```
manojgop: Better to return consistent types on success and failure. One option is to return a tuple with (signature, status). Also please update function comments accordingly

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -131,13 +131,25 @@ namespace tcf {
...
+            "Unsupported payload format found in the input");
+
         // Decrypt Encryption key
         ByteArray encrypted_session_key_bytes = HexStringToBinary(encrypted_session_key);
         session_key = enclaveData.decrypt_message(encrypted_session_key_bytes);
         ByteArray session_key_iv_bytes = HexStringToBinary(session_key_iv);
         JSON_Array* data_array = json_object_get_array(params_object, "inData");
-
         size_t count = json_array_get_count(data_array);
+        tcf::error::ThrowIf<tcf::error::ValueError>(count == 0, "Indata is empty");
```
manojgop: indata can be empty if work order doesn't need any in data

Ram-srini: As per the spec, it should have one or more items in indata. inData contains either a JWT of the specified data or an array of one or more Work Order inputs, e.g. state, message containing input parameters.

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
manju956: Done

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
manju956: Done

examples/common/python/utility/signature.py
manju956: Done

examples/common/python/utility/signature.py
manju956: Ok, i added a check to validate hex string

examples/common/python/utility/signature.py
manju956: Done

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -146,8 +145,23 @@ def __process_worker_register(self, worker_id, input_value_json, response):
...
+
 # ------------------------------------------------------------------------------------------------
+    def validate_input_worker_status(self, input_json):
+        '''
+        Validates worker status in the input request
+        '''
+        try:
+            worker_status = input_json["params"]["status"]
+            status = int(worker_status)
+            if status < WorkerStatus.ACTIVE.value or status > WorkerStatus.COMPROMISED.value:
```
Ram-srini: Don't compare the range using value. What if we add one more status type to enum?
Use type conversion something like below.
try:
            worker_status = input_json["params"]["status"]
            status = int(worker_status)
            wo_status = WorkerStatus(status)
            return True
except:
            logger.error("Invalid parameter format or value found in the request")
            return False

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -146,8 +145,23 @@ def __process_worker_register(self, worker_id, input_value_json, response):
...
+                return WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE
+        except:
+            logger.error("Invalid parameter format or value found in the request")
+            return WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE
```
Ram-srini: simplify it using True/False as returning value instead of workerror code.

examples/common/python/utility/signature.py
```diff
@@ -177,23 +177,28 @@ def calculate_datahash(self, data_objects):
     def generate_signature(self, hash, private_key):
```
Ram-srini: generate_signature() is used in work_order_params as well, changes are required there as well

manju956:
Ram raised another PR to address these changes. Hence closing this PR.
https://github.com/hyperledger-labs/trusted-compute-framework/pull/89

PR #92, opened 2019-09-25
Add client side signature verification check in following client implementations

Ram-srini:
Add client side signature verification check in following client implementations
echo client
heart disease eval client
generic client


examples/apps/echo/client/echo_client.py
```diff
@@ -237,10 +239,24 @@ def Main(args=None):
 		json.dumps(res, indent=4)
 	))
 	if "result" in res:
-		decrypted_res = utility.decrypted_response(
-			json.dumps(res), session_key, session_iv)
-		logger.info("\nDecrypted response:\n {}".format(decrypted_res))
+		sig_obj = signature.ClientSignature()
+		status = sig_obj.verify_signature(json.dumps(res), worker_obj)
```
manojgop: Can we change the input parameters of verify_signature() to use json dictionary instead of json string. This will avoid unnecessary conversion from json str to json dictionary and vice-versa. 
Also entire worker_obj need not be passed to the function.

examples/apps/echo/client/echo_client.py
```diff
@@ -237,10 +239,24 @@ def Main(args=None):
...
+		try:
+			if status == SignatureStatus.PASSED:
+				logger.info("Signature verification Successful")
+				decrypted_res = utility.decrypted_response(
+					json.dumps(res), session_key, session_iv)
```
manojgop: for function parameter, use json dictionary instead of json string

examples/apps/echo/client/echo_client.py
Ram-srini: worker_obj is passed because verify_signature is checking for hashing algorithm, signing algorithm and verification key from worker_obj thats it is passed as argument.

manojgop: Would it make more sense to pass just the required fields if its not too much of an ask ?

examples/apps/echo/client/echo_client.py
Ram-srini: Changed

examples/common/python/utility/signature.py
```diff
@@ -282,22 +282,25 @@ def generate_client_signature(self, input_json_str,
...
 
 #---------------------------------------------------------------------------------------------
-    def verify_signature(self,response_str,worker):
+    def verify_signature(self,input_json,worker):
         """
         Function to verify the signature received from the enclave
         Parameters:
-            - response_str is json payload returned by the Worker Service in response to successful workorder submit request as per TCF API 6.1.2 Work Order Result Payload
-            - worker is a worker object to store all the common details of worker as per TCF API 8.1 Common Data for All Worker Types
+            - response is dictionary contains payload returned by the
```
manojgop: Update parameters description for input_json

examples/common/python/utility/signature.py
```diff
@@ -282,22 +282,25 @@ def generate_client_signature(self, input_json_str,
         return input_json_str, SignatureStatus.PASSED
 
 #---------------------------------------------------------------------------------------------
-    def verify_signature(self,response_str,worker):
+    def verify_signature(self,input_json,worker):
```
manojgop: Do we need to pass entire worker object to thsi function ? can't we just pass worker's HashingAlgorithm and SigningAlgorithm ?

Ram-srini: Ideally verify_signature() it would take only 2 parameters verifying key and signature. It is checking additional thing by verifying the algorithms for hashing and signing. Do we really need that for verifying the signature?

manojgop: It's good to pass hashing and signing to verify_signature() function since it does hashing/signing operations and it uses some default hashing/signing algorithms.

PR #93, opened 2019-09-25
Use AES_GCM 256 bits for symmetric key encryption

manojgop:

### DIFF
--- a/tc/sgx/common/crypto/skenc.h
+++ b/tc/sgx/common/crypto/skenc.h
@@ -22,10 +22,10 @@
 namespace tcf {
 namespace crypto {
     namespace constants {
-        //***AES-GCM 128 for authenticated encryption
+        //***AES-GCM 256 for authenticated encryption
         const int IV_LEN = 12;       // AES-GCM IV length
-        const int SYM_KEY_LEN = 16;  // AES-GCM Key length. move to 32 for post-quantum security
-        const int TAG_LEN = 16;      // AES-GCM TAG length, move to 32 for post-quantum security
+        const int SYM_KEY_LEN = 32;  // AES-GCM Key length.
+        const int TAG_LEN = 16;      // AES-GCM TAG length.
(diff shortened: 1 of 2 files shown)


tc/sgx/common/crypto/skenc.h
```diff
@@ -22,9 +22,9 @@
 namespace tcf {
 namespace crypto {
     namespace constants {
-        //***AES-GCM 128 for authenticated encryption
+        //***AES-GCM 256 for authenticated encryption
         const int IV_LEN = 12;       // AES-GCM IV length
-        const int SYM_KEY_LEN = 16;  // AES-GCM Key length. move to 32 for post-quantum security
+        const int SYM_KEY_LEN = 32;  // AES-GCM Key length.
         const int TAG_LEN = 16;      // AES-GCM TAG length, move to 32 for post-quantum security
```
danintel: Remove the comment `move to 32 for post-quantum security`

manojgop: Done

manojgop:
I modified the commit to append authentication tag to cipher text to follow default convention by other crypto libraries.
In RFC 5116: An Interface and Algorithms for Authenticated Encryption the authentication tag is explicitly placed at the end of the ciphertext for GCM mode encryption

kubasiemion:
Nice, thanks for fixing that!

danintel:
What bothers me about this, and TCF in general, is the crypto algorithms are hard-coded. This goes against security best practices. For standards to exist long term, they cannot tie themselves to specific cryptographic algorithms. If they do, then they become quickly obsolete.  Now AES-256 is OK and will be OK for years or even decades. But we cannot rely on that.  Even if we support one option for now (AES-256), that will allow for migration in the future to other (perhaps preferred) crypto algorithms in itself.

However, the root problem is in the spec, not in the code. The change itself is OK.

PR #107, opened 2019-09-29
Allow Work orders without signature and verifying key

manju956:
- Since requester signature is not mandatory, work order requests
  without signature and verifying key should be processed.
- Added command line option to enable requester signature.
- Changes to allow empty/no resultUri, notifyUri, workerEncryptionKey and dataEncryptionAlgorithm are added.

### DIFF
--- a/examples/apps/echo/client/echo_client.py
+++ b/examples/apps/echo/client/echo_client.py
@@ -50,6 +50,7 @@ def ParseCommandLine(args) :
 	global message
 	global config
 	global off_chain
+	global requester_signature
 
 	parser = argparse.ArgumentParser()
 	use_service = parser.add_mutually_exclusive_group()
@@ -71,6 +72,9 @@ def ParseCommandLine(args) :
 	parser.add_argument("-m", "--message", 
 		help='text message to be included in the JSON request payload', 
 		type=str)
+	parser.add_argument("-rs", "--requester_signature",
+		help="Enable requester signature for work order requests",
+		action="store_true")
 
 	options = parser.parse_args(args)
 
@@ -103,6 +107,8 @@ def ParseCommandLine(args) :
 
 	if options.off_chain:
 		off_chain = True
+
+	requester_signature = options.requester_signature
 
 	worker_id = options.worker_id
 	message = options.message
@@ -203,12 +209,16 @@ def Main(args=None):
 	# Add worker input data
 	wo_params.add_in_data(message)
 
-	# Sign work order
-	private_key = utility.generate_signing_keys()
+	# Encrypt work order request hash
 	wo_params.add_encrypted_request_hash()
-	if wo_params.add_requester_signature(private_key) == False:
-		logger.info("Work order request signing failed\n")
-		sys.exit(1)
+
+	if requester_signature:
+		private_key = utility.generate_signing_keys()
+		# Add requester signature and requester verifying_key
+		if wo_params.add_requester_signature(private_key) == False:
+			logger.info("Work order request signing failed")
+			exit(1)
+
 	# Submit work order
 	logger.info("Work order submit request : %s, \n \n ",
         wo_params.to_string())
--- a/examples/apps/heart_disease_eval/client/heart_gui.py
+++ b/examples/apps/heart_disease_eval/client/heart_gui.py
@@ -232,11 +232,15 @@ class resultWindow(tk.Toplevel):
 		)
 		wo_params.add_in_data(message)
 
-		private_key = utility.generate_signing_keys()
 		wo_params.add_encrypted_request_hash()
-		if wo_params.add_requester_signature(private_key) == False:
-			logger.info("Work order request signing failed\n")
-			sys.exit(1)
+
+		if requester_signature:
+			private_key = utility.generate_signing_keys()
+			# Add requester signature and requester verifying_key
+			if wo_params.add_requester_signature(private_key) == False:
+				logger.info("Work order request signing failed")
+				exit(1)
+
 		# Set text for JSON sidebar
 		req_id = 51
 		self.request_json = wo_params.to_string()
@@ -423,6 +427,7 @@ def ParseCommandLine(args) :
 	global verbose
 	global config
 	global off_chain
+	global requester_signature
 
 	parser = argparse.ArgumentParser()
 	use_service = parser.add_mutually_exclusive_group()
@@ -443,6 +448,9 @@ def ParseCommandLine(args) :
 		type=str)
 	parser.add_argument("-v", "--verbose", 
 		help="increase output verbosity", 
+		action="store_true")
+	parser.add_argument("-rs", "--requester_signature",
+		help="Enable requester signature for work order requests",
 		action="store_true")
 
 	options = parser.parse_args(args)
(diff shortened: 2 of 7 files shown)


tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -363,14 +367,14 @@ namespace tcf {
                 out_data.ComputeHashString();
             } else {
                 // Assign data_encryption_key, iv, data_iv and encrypted_data_encryption_key
-                // params with corresponding values from inData
+                // params with corresponging values from inData
                 ByteArray encryption_key = data_items_in[i].GetEncryptionKey();
                 std::string iv = data_items_in[i].GetIv();
-                std::string encrypted_data_encryption_key =
+                std::string encrypted_data_encrption_key =
```
danintel: s/encrption/encryption/
.... Also other places

manju956: Done

examples/apps/echo/client/echo_client.py
```diff
@@ -201,12 +201,10 @@ def Main(args=None):
 	# Add worker input data
 	wo_params.add_in_data(message)
 
-	# Sign work order
+	# Encrypt work order request hash
 	private_key = utility.generate_signing_keys()
```
manojgop: Do client need to generate private key if it's not used for signing

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -61,10 +61,12 @@ namespace tcf {
             "payloadFormat",
             "invalid request; failed to retrieve payload format");
 
+        // verifyingKey is optional field. Hence don't throw exception
```
manojgop: For better clarity, this could be renamed as requesterVerifyingKey. Also add a comment this parameter is not mentioned in spec and added to verify requesterSignature

tests/test_wo_with_no_signature.py
```diff
@@ -0,0 +1,257 @@
+#! /usr/bin/env python3
```
manojgop: Add a test case for using requesterSignature

examples/apps/echo/client/echo_client.py
manju956: Its not required, will delete it

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
manju956: Ok, i will add a comment

tests/test_wo_with_no_signature.py
manju956: Done

examples/apps/echo/client/echo_client.py
```diff
@@ -201,12 +201,9 @@ def Main(args=None):
 	# Add worker input data
 	wo_params.add_in_data(message)
 
-	# Sign work order
-	private_key = utility.generate_signing_keys()
+	# Encrypt work order request hash
 	wo_params.add_encrypted_request_hash()
-	if wo_params.add_requester_signature(private_key) == False:
```
Ram-srini: Why do we need to remove signature part? Signature is optional doesn't mean to remove it everywhere. I think we should keep it. Since it is reference client and if some use this as reference then we should provide all option to add signature and verify signature.

examples/apps/heart_disease_eval/client/heart_gui.py
```diff
@@ -230,11 +230,8 @@ def evaluate(self, message):
 		)
 		wo_params.add_in_data(message)
 
-		private_key = utility.generate_signing_keys()
 		wo_params.add_encrypted_request_hash()
-		if wo_params.add_requester_signature(private_key) == False:
-			logger.info("Work order request signing failed\n")
-			sys.exit(1)
+
```
Ram-srini: Please re base your PR

manju956: Done

examples/apps/echo/client/echo_client.py
manojgop: The issue with adding requesterSignature in request is we are also passing verifyingKey in JSON request to worker enclave which is outside the spec. So this may lead to confusion especially when 3rd parties are writing client applications in different language other than Python. My understanding is @manju956 will write a test case for requesterSignature as a reference code.

manojgop: The issue with adding requesterSignature in request is we are also passing verifyingKey in JSON request to worker enclave which is outside the spec. So this may lead to confusion especially when 3rd parties are writing client applications in different language other than Python. My understanding is @manju956 will write a test case for requesterSignature as a reference code.

manju956: test case which includes requesterSignature and verifyingKey is already included in this PR.

tests/test_wo_with_no_signature.py
```diff
@@ -0,0 +1,256 @@
...
+		type=str)
+	use_service.add_argument("-o", "--off-chain",
+		help="skip URI lookup and use the registry in the config file",
+		action="store_true")
+	parser.add_argument("-w", "--worker-id",
+		help="skip worker lookup and retrieve specified worker",
+		type=str)
+	parser.add_argument("-m", "--message",
+		help='text message to be included in the JSON request payload',
+		type=str)
```
Ram-srini: Add -s command line argument to existing echo client implementation. If -s passed then  add requesterSignature otherwise no signature. No need to add entire python code as test case, call "echo_client.py -s" option in run_tests.sh script.

tests/test_wo_with_signature.py
```diff
@@ -0,0 +1,262 @@
...
+	private_key = utility.generate_signing_keys()
+	if wo_params.add_requester_signature(private_key) == False:
+		logger.error("Work order request signing failed")
+		sys.exit(1)
+
+	# Submit work order
+	logger.info("Work order submit request : %s, \n \n ",
+        wo_params.to_string())
+	work_order_instance = direct_jrpc.create_work_order(
+		config
```
Ram-srini: Duplicate code - call echo client without signing option in run_tests.sh

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -126,10 +131,12 @@ namespace tcf {
             "encryptedRequestHash",
             "invalid request; failed to retrieve encrypted request hash");
 
+        // requesterSignature is optional field. Hence don't throw exception
+        // if param is not there or empty in the request.
         requester_signature = GetJsonStr(
             params_object,
```
Ram-srini: Could you please make changes to below optional fields
1. resultUri
2. notifyUri
3. workerEncryptionKey
4. dataEncryptionAlgorithm

manju956: Done

examples/apps/echo/client/echo_client.py
manju956: I have added command line option to enable requester signature

tests/test_wo_with_no_signature.py
manju956: I have added -r option to command line argument to enable requester signature. 
Also, i will remove my test cases

tests/test_wo_with_signature.py
manju956: Sure, will delete it

PR #118, opened 2019-10-04
Verify data hash of input data

manju956:
- Verify data hash of input data if present in work order request.
- Also, data hash of work order response is verified at the client.

### DIFF
--- a/examples/apps/echo/client/echo_client.py
+++ b/examples/apps/echo/client/echo_client.py
@@ -35,6 +35,7 @@ from connectors.direct.direct_json_rpc_api_connector \
 import crypto.crypto as crypto
 from error_code.error_status import WorkOrderStatus
 import utility.signature as signature
+import utility.hex_utils as hex_utils
 from error_code.error_status import SignatureStatus
 
 # Remove duplicate loggers
@@ -51,6 +52,7 @@ def ParseCommandLine(args) :
 	global config
 	global off_chain
 	global requester_signature
+	global input_data_hash
 
 	parser = argparse.ArgumentParser()
 	use_service = parser.add_mutually_exclusive_group()
@@ -72,8 +74,11 @@ def ParseCommandLine(args) :
 	parser.add_argument("-m", "--message", 
 		help='text message to be included in the JSON request payload', 
 		type=str)
-	parser.add_argument("-rs", "--requester_signature",
+	parser.add_argument("-rs", "--requester-signature",
 		help="Enable requester signature for work order requests",
+		action="store_true")
+	parser.add_argument("-dh", "--data-hash",
+		help="Enable input data hash for work order requests",
 		action="store_true")
 
 	options = parser.parse_args(args)
@@ -109,7 +114,7 @@ def ParseCommandLine(args) :
 		off_chain = True
 
 	requester_signature = options.requester_signature
-
+	input_data_hash = options.data_hash
 	worker_id = options.worker_id
 	message = options.message
 	if options.message == None or options.message == "":
@@ -198,6 +203,7 @@ def Main(args=None):
 	session_iv = utility.generate_iv()
 	session_key = utility.generate_key()
 	requester_nonce = secrets.token_hex(16)
+
 	# Create work order
 	wo_params = WorkOrderParams(
 		work_order_id, worker_id, workload_id, requester_id,
@@ -207,7 +213,14 @@ def Main(args=None):
 		data_encryption_algorithm="AES-GCM-256"
 	)
 	# Add worker input data
-	wo_params.add_in_data(message)
+	if input_data_hash:
+		# Compute data hash for data params inData
+		data_hash = utility.compute_data_hash(message)
+		# Convert data_hash to hex
+		data_hash = hex_utils.byte_array_to_hex_str(data_hash)
+		wo_params.add_in_data(message, data_hash)
+	else:
+		wo_params.add_in_data(message)
 
 	# Encrypt work order request hash
 	wo_params.add_encrypted_request_hash()
@@ -257,6 +270,12 @@ def Main(args=None):
 				decrypted_res = utility.decrypted_response(
 					res, session_key, session_iv)
 				logger.info("\nDecrypted response:\n {}".format(decrypted_res))
+				if input_data_hash:
+					decrypted_data = decrypted_res[0]["data"]
+					data_hash_in_resp = (decrypted_res[0]["dataHash"]).upper()
+					# Verify data hash in response
+					if utility.verify_data_hash(decrypted_data, data_hash_in_resp) == False:
+						sys.exit(1)
 			else:
 				logger.info("Signature verification Failed")
 				sys.exit(1)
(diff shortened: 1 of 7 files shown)


manju956:
> I'll approve this, but I still do not like that the hash algorithms are hard-coded to SHA-256.
> But this is a defect of the TCF specification, not your code.
> 
> Also, you added a test. Shouldn't the test be invoked somewhere too for, say, CI?

Yes Dan, it makes sense to trigger the test case from run_tests.sh script which will be run as part of CI.
I will raise a seperate PR to include other test cases from run_tests.sh as well.

tests/test_verify_data_hash.py
```diff
@@ -0,0 +1,278 @@
+#! /usr/bin/env python3
```
manojgop: Better to add this feature in one of the existing clients (echo or generic client) and call it from run_tests.sh instead of duplicating the code

manju956: Done

tc/sgx/trusted_worker_manager/enclave/work_order_data_handler.cpp
```diff
@@ -131,9 +131,18 @@ namespace tcf {
...
     }
 
-    void WorkOrderDataHandler::VerifyInputHash(ByteArray input_data,
-		                               ByteArray input_hash) {
-        // Do nothing at the phase 1
+    tcf_err_t WorkOrderDataHandler::VerifyInputHash(ByteArray input_data,
+                                                    ByteArray input_hash) {
+        tcf_err_t verify_status = TCF_SUCCESS;
+        ByteArray hash = tcf::crypto::ComputeMessageHash(input_data);
+        if (std::equal(hash.begin(), hash.end(), input_hash.begin())) {
```
arsulegai: Question: `input_hash` is coming from the external source. `hash` is internally calculated, will the equal method take care if the `input_hash` is of smaller length than `hash`?

examples/apps/echo/client/echo_client.py
```diff
@@ -207,7 +212,14 @@ def Main(args=None):
 		data_encryption_algorithm="AES-GCM-256"
 	)
 	# Add worker input data
-	wo_params.add_in_data(message)
+	if input_data_hash:
+		# Compute data hash for data params inData
+		data_hash = utility.compute_data_hash_b64(message)
```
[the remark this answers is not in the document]
manju956: earlier version of function was returning b64 encoded value. forgot to remove b64 from the function name after removing b64 encoding line inside function. I will correct the function name.

tools/run_tests.sh
```diff
@@ -23,6 +23,7 @@ fi
```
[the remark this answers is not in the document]
manju956: Done

tc/sgx/trusted_worker_manager/enclave/work_order_data_handler.cpp
```diff
@@ -131,9 +131,18 @@ namespace tcf {
 
     }
 
-    void WorkOrderDataHandler::VerifyInputHash(ByteArray input_data,
-		                               ByteArray input_hash) {
-        // Do nothing at the phase 1
+    tcf_err_t WorkOrderDataHandler::VerifyInputHash(ByteArray input_data,
```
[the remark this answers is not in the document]
manju956: I will add a check to handle the return value from the function.

tc/sgx/trusted_worker_manager/enclave/work_order_data_handler.cpp
manju956: Yes, std::equal function takes care if input_hash is of smaller length

manju956:
> Another request: Please create two commits for this change. One for the example implementation and the other for the enclave code.

All the changes are related to data hash verification. I am not sure why 2 commits are needed.

arsulegai:
> > Another request: Please create two commits for this change. One for the example implementation and the other for the enclave code.
> 
> All the changes are related to data hash verification. I am not sure why 2 commits are needed.

Smaller related commits make it easier for maintenance.
1. Commit 1: Feature enabling the input data hash verification (the enclave part). This is applicable to all the client requests. This will function as is irrespective of which client is sending the request.
2. Commit 2: Make use of this feature for the echo client. This commit can be referenced by others to add similar feature in their client.

examples/apps/echo/client/echo_client.py
```diff
@@ -207,7 +212,14 @@ def Main(args=None):
 		data_encryption_algorithm="AES-GCM-256"
 	)
 	# Add worker input data
-	wo_params.add_in_data(message)
+	if input_data_hash:
+		# Compute data hash for data params inData
+		data_hash = utility.compute_data_hash(message)
+		# Convert data_hash to hex
+		data_hash = ''.join(format(i, '02x') for i in data_hash)
```
manojgop: Don't we have a utility to convert to hex ? If not, can we add it to utility and use it.

examples/apps/echo/client/echo_client.py
```diff
@@ -257,6 +269,20 @@ def Main(args=None):
...
+					# Convert data_hash to hex
+					computed_data_hash = ''.join(format(i, '02x') \
+						for i in computed_data_hash).upper()
+					data_hash_in_resp = (decrypted_res[0]["dataHash"]).upper()
+					if computed_data_hash == data_hash_in_resp:
+						logger.info("Computed hash of decrypted data matched "\
+							"with data hash in the response")
+					else:
+						logger.error("Computed hash of decrypted data does not"\
+							"match with data hash in the response")
```
manojgop: Would you like to move data hash verification logic to a function for better readability ? May be check the return value of the function and exit if verification fails.

examples/apps/echo/client/echo_client.py
manojgop: It makes sense to write a utility function to convert Byte Array to hex string. I can see the same logic duplicated in this file.

examples/apps/echo/client/echo_client.py
manju956: Could be done, it doesn't much reduce the lines (lines 275 and 277 can be avoided.).  We still need to get decrypted data and hash value from the response and still need to have validation in place.

examples/apps/echo/client/echo_client.py
manju956: Ok

examples/apps/echo/client/echo_client.py
manojgop: The function can return a boolean. You can also move the logging inside this function which is fine as this will be an internal function.

manju956: Done

examples/common/python/utility/utility.py
```diff
@@ -245,7 +255,25 @@ def decrypted_response(input_json, session_key, session_iv, data_key=None, data_
     return input_json_params['outData']
 
 #---------------------------------------------------------------------------------------------
+def verify_data_hash(msg, data_hash):
```
manojgop: For now it's fine to keep this function. But we need to revisit this utility.py and see how to organize it better by moving some functions outside this file as a separate PR. For example, encryption related utility functions can be move to encrypt_utils.py,  file related utility functiosn to file_utils.py and so on/

PR #123, opened 2019-10-08
adding contract for compliency with v1.1 of the spec

Amxx:
Replaces #72

### DIFF
--- a/examples/common/python/connectors/ethereum/contracts/proxy-model/WorkerRegistry.sol
+++ b/examples/common/python/connectors/ethereum/contracts/proxy-model/WorkerRegistry.sol
@@ -1,0 +1,142 @@
+/* Copyright 2019 iExec Blockchain Tech
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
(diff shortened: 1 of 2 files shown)


examples/common/python/connectors/ethereum/contracts/proxy-model/WorkerRegistry.sol
```diff
@@ -0,0 +1,142 @@
...
+		uint256   workerType;
+		bytes32   organizationId;
+		bytes32[] appTypeIds;
+		string    details;
+	}
+
+	// WorkerID → Worker
+	mapping(bytes32 => Worker) private m_workers;
+
+	// workerType → organizationId → appTypeId → WorkerID[]
```
danintel: From #72:
s/appTypeId/appTypeIds/

(I see no appTypeId here, only appTypeIds)

Amxx: `appTypeIds` are arrays of `appTypeId`
to make worker searcheable by invididual appTypeId, we have to split the set of supported appTypeIds and enter as many entries in this multidimensional mapping.

`workerRegister`'s code should show that clearly

PR #126, opened 2019-10-08
Set SGX_MODE Default

danintel:
Set SGX_MODE default to `SIM`. This makes SGX_MODE optional.


examples/enclave_manager/setup.py
```diff
@@ -33,7 +33,7 @@
 module_path = 'tc/sgx/trusted_worker_manager/enclave_wrapper'
 module_src_path = os.path.join(tcf_root_dir, module_path)
 
-sgx_mode_env = os.environ.get('SGX_MODE', None)
+sgx_mode_env = os.environ.get('SGX_MODE', 'SIM')
```
manju956: Could you please add similar changes in setup.py in below path ?
examples/common/python/setup.py

danintel: I already do that.  See the change above in `examples/common/python/setup.py`:
```
- sgx_mode_env = os.environ.get('SGX_MODE', None)
+ sgx_mode_env = os.environ.get('SGX_MODE', 'SIM')
```

danintel:
@manju956: look again. I set `SGX_MODE` in both Python `setup.py` files, 
`examples/common/python/setup.py` and `examples/enclave_manager/setup.py`

manju956:
> @manju956 : look again. I set `SGX_MODE` in both Python `setup.py` files,
> `examples/common/python/setup.py` and `examples/enclave_manager/setup.py`

My bad, missed that file.

docker-compose.yaml
```diff
@@ -34,6 +34,8 @@ services:
       - /tmp/.X11-unix:/tmp/.X11-unix
       - ~/.Xauthority:/root/.Xauthority
     working_dir: "/project/TrustedComputeFramework/tools/build"
+    environment:
```
arsulegai: Is this required, given that the setup.py, cmake/make files are assuming the default value to be SIM?

danintel: @arsulegai. No this is not required. It makes it obvious for anyone using this script that it is for SGX simulator mode use only.

PR #127, opened 2019-10-09
Avoid hashing twice for signing data.

manojgop:
SignMessage() function is hashing the data before signing.
Client is expected to hash the data as mentioned in Trusted computing
spec. So no need to hash the data again before signing.

### DIFF
--- a/tc/sgx/common/crypto/sig_private_key.cpp
+++ b/tc/sgx/common/crypto/sig_private_key.cpp
@@ -206,16 +206,15 @@ std::string pcrypto::sig::PrivateKey::Serialize() const {
 }  // pcrypto::sig::PrivateKey::Serialize
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-// Computes SHA256 hash of message.data(), signs with ECDSA privkey and
-// returns ByteArray containing raw binary data
+// Signs hashMessage.data() with ECDSA privkey and returns ByteArray
+// containing raw binary data. It's expected that caller of this function
+// passes hash value of original message to this function for signing.
 // throws RuntimeError
-ByteArray pcrypto::sig::PrivateKey::SignMessage(const ByteArray& message) const {
-    unsigned char hash[SHA256_DIGEST_LENGTH];
-    // Hash
-    SHA256((const unsigned char*)message.data(), message.size(), hash);
-
-    // Then Sign
-    ECDSA_SIG_ptr sig(ECDSA_do_sign((const unsigned char*)hash, SHA256_DIGEST_LENGTH, private_key_),
+ByteArray pcrypto::sig::PrivateKey::SignMessage(
+    const ByteArray& hashMessage) const {
+    // Sign
+    ECDSA_SIG_ptr sig(ECDSA_do_sign((const unsigned char*)hashMessage.data(),
+        hashMessage.size(), private_key_),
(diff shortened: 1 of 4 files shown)


tc/sgx/common/crypto/sig_public_key.cpp
```diff
@@ -296,22 +296,20 @@ std::string pcrypto::sig::PublicKey::SerializeXYToHex() const {
 }  // SerializeXYToHex
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-// Verifies SHA256 ECDSA signature of message
+// Verifies ECDSA signature of message
 // input signature ByteArray contains raw binary data
```
manju956: Can we change raw binary data to hash of binary data?

tc/sgx/common/crypto/sig_private_key.cpp
```diff
@@ -206,16 +206,13 @@ std::string pcrypto::sig::PrivateKey::Serialize() const {
 }  // pcrypto::sig::PrivateKey::Serialize
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-// Computes SHA256 hash of message.data(), signs with ECDSA privkey and
-// returns ByteArray containing raw binary data
+// Signs message.data() with ECDSA privkey and returns ByteArray
+// containing raw binary data
```
manju956: Can we change raw binary data -> signature bytes ?

tc/sgx/common/crypto/sig_public_key.cpp
manojgop: As far this function is concerned, it treats input as raw binary data. It need not be aware of the source of data (hashed or not).

tc/sgx/common/crypto/sig_private_key.cpp
manojgop: It loks fine to me. Signs and returns ByteArray containing raw binary data

manju956: ok

tc/sgx/common/crypto/sig_public_key.cpp
manju956: ok

tc/sgx/common/crypto/sig_private_key.cpp
```diff
@@ -206,16 +206,13 @@ std::string pcrypto::sig::PrivateKey::Serialize() const {
 }  // pcrypto::sig::PrivateKey::Serialize
```
arsulegai: ECDSA_do_sign and ECDSA_verify_sign functions do not compute the hash by itself. Generally the signing utility function accepts hashing algorithm, message in raw and the key to compute the signature output or to verify the signature.
Ref: https://github.com/openssl/openssl/blob/master/crypto/ec/ecdsa_sign.c#L14
https://github.com/openssl/openssl/blob/6f02932edba62186a6866e8c9f0f0714674f6bab/crypto/ec/ecdsa_vrf.c#L20

I guess the change has to be done in Python and not in C++.

Please add back comments as well, the name of the hashing algorithm which is always SHA256 for this sign/verify functionality.

manojgop: The parameter name message is confusing with new changes done in this PR. I'll change it to hashMessage

PR #130, opened 2019-10-11
Build refactor

manju956:
- Downloading pip packages is moved from Makefile to Docker image
- Removed virtual environment in docker build
- Python egg packaging is replaced to wheel packaging

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -100,9 +100,12 @@ The steps below will set up a Python virtual environment to run TCF.
    Refer to the [PREREQUISITES document](PREREQUISITES.md)
    for more details on the above variables
 
-5. Build the Python virtual environment and install TCF components into it:
+5. Create Python virtual environment, Build and Install TCF
+   components into it:
    ```
    cd $TCF_HOME/tools/build
+   # Create virtual environment directory with name _dev
+   python3 -m venv _dev
    sudo make clean
    make
    ```
@@ -154,6 +157,10 @@ Follow these steps to run the `Demo.py` testcase:
 7. If you wish to exit the TCF program, press `y` and `Enter` at Terminal 1
    for standalone builds.
    For Docker-based builds, press `Ctrl-c`
+8. For standalone mode, delete virtual environment
+   ```
+   rm -rf $TCF_HOME/tools/build/_dev/
+   ```
 
 A GUI is also available to run this demo.
 See [examples/apps/heart_disease_eval](examples/apps/heart_disease_eval)
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -49,9 +49,13 @@ RUN apt-get update \
     cmake \
     protobuf-compiler \
     libprotobuf-dev \
-    python \
     python3-dev \
     virtualenv \
+    python3-pip \
+    python3-toml \
+    python3-requests \
+    python3-colorlog \
+    python3-twisted \
     swig \
     software-properties-common \
     curl \
@@ -67,6 +71,13 @@ RUN apt-get update \
  && apt-get -y -q upgrade \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
+
+# Install setuptools, py-solc and web3 packages using pip because
+# these are not available in apt repository.
+RUN pip3 install --upgrade setuptools json-rpc py-solc web3
+
+# Make Python3 default
+RUN ln -s /usr/bin/python3 /usr/bin/python
 
 # SGX common library and SDK are installed in /opt/intel directory.
 # Installation of SGX libsgx-common packages requires /etc/init directory. In docker image this directory doesn't exist.
--- a/examples/common/python/Makefile
+++ b/examples/common/python/Makefile
@@ -12,18 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/'}
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
 MOD_VERSION=${shell ../../../bin/get_version}
 
-EGG_FILE=dist/tcf_examples_common-${MOD_VERSION}-py${PY_VERSION}-linux-x86_64.egg
+# Format of wheel package is <pkg_name>-{python-tag}-{abi-tag}-{platform}
+WHEEL_FILE=dist/tcf_examples_common-${MOD_VERSION}-cp${PY_VERSION}-cp${PY_VERSION}m-linux_x86_64.whl
 SOURCE_DIR=$(shell pwd)
 
 
-all : $(EGG_FILE)
+all : $(WHEEL_FILE)
 
-$(EGG_FILE) : build_ext
+$(WHEEL_FILE) : build_ext
 	@echo Build Distribution
-	python3 setup.py bdist_egg
+	python3 setup.py bdist_wheel
 	
 build_ext :
 	@echo Build build_ext
@@ -34,10 +35,11 @@ build :
 	cd $@ 
 
 install:
-	@echo INSTALLING EGG FILE =================
-	easy_install $(EGG_FILE)
+	@echo INSTALLING WHEEL FILE =================
+	pip3 install $(WHEEL_FILE)
 
 clean:
+	pip3 uninstall --yes $(WHEEL_FILE)
 	rm -rf build deps dist *.egg-info
 	rm -f crypto/crypto.py crypto/crypto_wrap.cpp
 	rm -f verify_report/verify_report.py verify_report/verify_report_wrap.cpp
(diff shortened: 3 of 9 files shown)


tools/build/Makefile
```diff
@@ -25,9 +25,16 @@ PYTHON_DIR=$(DSTDIR)/lib/python$(PY_VERSION)/site-packages/
 
 all: environment template enclavekey packages
 
+no_venv: template enclavekey packages_no_venv
```
danintel: I would add a 1 line comment that this is the same as `make all` without setting up a Python virtual environment.

manojgop: You could consider conditionally building a target and it's dependencies instead of adding multiple targets like no_env, clean_pip_packages, packages_no_env
for example,
ALL_TARGETS = template enclavekey packages
ifeq ($(STAND_ALONE_BUILD),1)
    ALL_TARGETS += environment
endif
all: $(ALL_TARGETS)

clean :
     ifeq ($(STAND_ALONE_BUILD),1)
          rm -rf /usr/lib/python3/dist-packages/tcf*
     else
          rm -rf $(DSTDIR)
      endif
      $(abspath $(SRCDIR)/../clean.sh)

docker/Dockerfile.tcf-dev
```diff
@@ -68,6 +73,12 @@ RUN apt-get update \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
 
+# Install setuptools, py-solc and web3 packages
```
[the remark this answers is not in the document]
arsulegai: This shall be added to the documentation as well.

docker/Dockerfile.tcf-dev
```diff
@@ -49,9 +49,14 @@ RUN apt-get update \
...
     libprotobuf-dev \
-    python \
     python3-dev \
     virtualenv \
+    python3-pip \
+    python3-toml \
+    python3-requests \
+    python3-colorlog \
+    python3-twisted \
+    python3-protobuf \
```
Ram-srini: Are we using protobuf anywhere in our project? Remove it if we are not using it.

docker/Dockerfile.tcf-dev
```diff
@@ -49,9 +49,14 @@ RUN apt-get update \
     cmake \
     protobuf-compiler \
     libprotobuf-dev \
-    python \
     python3-dev \
     virtualenv \
+    python3-pip \
```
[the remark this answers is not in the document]
manju956: Done

docker/Dockerfile.tcf-dev
manju956: I will remove it

docker/Dockerfile.tcf-dev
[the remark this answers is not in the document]
manju956: Done

tools/build/Makefile
manju956: I moved installation instruction of pip packages to pre-requisites section in documentation and Make file will not download any packages. In the cleanup section, source code related packages are deleted either from standard location as in the case of docker or virtual environment in the standalone mode.

arsulegai: Ok. Let's not call standalone or docker now onwards. Let's call it install and uninstall irrespective of which machine it is :)

BUILD.md
```diff
@@ -100,9 +100,12 @@ The steps below will set up a Python virtual environment to run TCF.
    Refer to the [PREREQUISITES document](PREREQUISITES.md)
    for more details on the above variables
 
-5. Build the Python virtual environment and install TCF components into it:
+5. Create Python virtual environment, Build and Install install TCF
```
manojgop: Type . Remove duplicate install "Install install"

examples/common/python/Makefile
```diff
@@ -12,18 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/'}
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
 MOD_VERSION=${shell ../../../bin/get_version}
 
-EGG_FILE=dist/tcf_examples_common-${MOD_VERSION}-py${PY_VERSION}-linux-x86_64.egg
+# Format of wheel package is <pkg_name>-{python-tag}-{abi-tag}-{platform}
+WHEEL_FILE=dist/tcf_examples_common-${MOD_VERSION}-cp${PY_VERSION}-cp${PY_VERSION}m-linux_x86_64.whl
```
manojgop: Do we need "cp${PY_VERSION}-cp${PY_VERSION}". Is this something to do with whl file naming convention ?

examples/common/python/Makefile
```diff
@@ -34,10 +35,11 @@ build :
...
 
 install:
-	@echo INSTALLING EGG FILE =================
-	easy_install $(EGG_FILE)
+	@echo INSTALLING WHEEL FILE =================
+	pip3 install $(WHEEL_FILE)
 
 clean:
+	pip3 uninstall --yes $(WHEEL_FILE)
 	rm -rf build deps dist *.egg-info
```
manojgop: I guess we can remove this line since egg is not used ?

tools/build/Makefile
```diff
@@ -23,31 +23,13 @@ PKGGEN=$(abspath $(SRCDIR)/../rebuild.sh)
 PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/'}
 PYTHON_DIR=$(DSTDIR)/lib/python$(PY_VERSION)/site-packages/
 
-all: environment template enclavekey packages
+all: template enclavekey packages
```
manojgop: You could remove environment from .PHONY target

examples/enclave_manager/Makefile
```diff
@@ -15,20 +15,19 @@
 # Set SGX_MODE default (which can be overridden with an environment variable):
 SGX_MODE?=SIM
 
-PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/'}
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
```
manojgop: What is "tr -d" for. I don;t see this option in PY_VERSION in "tools/build/Makefile" below

examples/enclave_manager/Makefile
```diff
@@ -15,20 +15,19 @@
...
 MOD_VERSION=${shell ../../bin/get_version}
 
-EGG_FILE=dist/tcf_enclave_manager-${MOD_VERSION}-py${PY_VERSION}-linux-x86_64.egg
+WHEEL_FILE=dist/tcf_enclave_manager-${MOD_VERSION}-cp${PY_VERSION}-cp${PY_VERSION}m-linux_x86_64.whl
```
manojgop: We have one more egg file in "examples/common/python/Makefile". Are we planning to replace that with whl file in a separate PR ?

BUILD.md
manojgop: Did you get a chance to try the standalone build with the changes in this PR

manju956: Will correct the typo.
I have tried to install virtualenv on standalone build and tried pip install/uninstall of tcf python packages and it works fine.
The whole set of instructions for standalone build is not tried out.

examples/common/python/Makefile
manju956: Yes, this is the naming convention followed by wheel package.

examples/common/python/Makefile
manju956: All these folders are still created and egg-info is created by pip. pip identifies whether a package is installed or not based on the existance of "egg-info" folder. It has nothing to do with egg packaging.

tools/build/Makefile
manju956: Done

examples/enclave_manager/Makefile
manju956: Its already replaced in this PR.

examples/enclave_manager/Makefile
manju956: This is needed because in wheel package naming convention, python tag doesn't have dot.
for eg: cp36 for python version 3.6

examples/common/python/Makefile
manojgop: ok. I was reading in below link which mentions -> Wheel uses PEP376-compliant .dist-info directories. Egg used .egg-info.
https://packaging.python.org/discussions/wheel-vs-egg/

danintel:
This PR added a new requirement---a new pip package--but did not document this fact.

I am fixing this undocumented change with this PR:
https://github.com/hyperledger-labs/trusted-compute-framework/pull/130

examples/enclave_manager/Makefile
danintel: This change requires an undocumented extra step:
`pip3 install wheel`
This was found from here:  https://github.com/tensorflow/tensorflow/issues/348

The correction to this regression is at #144

danintel:
This PR causes standalone builds to fail after building with Docker.  The solution is to enhance/update `scripts/clean.sh` to remove the root-owned build files created with the Docker build. See JIRA Ticket TCFH-502.

PR #131, opened 2019-10-11
Error response should have error keyword

manju956:
- Modify response to have error keyword.
- Skip signature verification if the response has error keyword.

### DIFF
--- a/tests/Demo.py
+++ b/tests/Demo.py
@@ -27,7 +27,7 @@ import worker.worker_details as worker
 from shared_kv.shared_kv_interface import KvStorage
 import utility.utility as enclave_helper
 import utility.file_utils as futils
-from error_code.error_status import SignatureStatus
+from error_code.error_status import SignatureStatus, WorkOrderStatus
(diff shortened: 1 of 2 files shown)


tests/Demo.py
```diff
@@ -120,7 +120,7 @@ def LocalMain(config) :
 
             # Polling for the "WorkOrderGetResult" and break when you get the result
             while("WorkOrderGetResult" in input_json_str1 and "result" not in response):
-                if response["error"]["code"] == 9:
+                if response["error"]["code"] != 5:
```
arsulegai: If you don't mind, could you please use a constant variable or add a comment for the meaning of the value 5?

danintel: This probably means pending.  I would just define a constant at the top of this file until we have a file with all the response codes defined.  Something like:
`PENDING_RESPONSE = 5`

manju956: I will make use of WorkOrderStatus enum.

PR #106, opened 2019-09-29
introduce library json-rpc to ease unwrapping requests and wrapping responses in tcs_worker_order_handler

sammyne:
Manually coding the process of unwrapping requests and wrapping responses in JSON-RPC is error-prone and tedious. So I make use of json-rpc library to simplify this process

Signed-off-by: Sammy <lixiangmin01@baidu.com>

PR #132, opened 2019-10-12
ease JSON-RPC requests handling for tcs_worker_registry_handler with json-rpc

sammyne:
1. this follows [PR #106](https://github.com/hyperledger-labs/trusted-compute-framework/pull/106)
2. replace the manually unwrapping requests and wrapping response with json-rpc

### DIFF
--- a/examples/common/python/connectors/direct/tcs_listener/tcs_listener.py
+++ b/examples/common/python/connectors/direct/tcs_listener/tcs_listener.py
@@ -84,6 +84,12 @@ class TCSListener(resource.Resource):
 
         self.dispatcher = Dispatcher()
         rpc_methods = [
+            self.worker_registry_handler.WorkerLookUp,
+            self.worker_registry_handler.WorkerLookUpNext,
+            self.worker_registry_handler.WorkerRegister,
+            self.worker_registry_handler.WorkerSetStatus,
+            self.worker_registry_handler.WorkerRetrieve,
+            self.worker_registry_handler.WorkerUpdate,
             self.workorder_handler.WorkOrderSubmit,
             self.workorder_handler.WorkOrderGetResult,
         ]
@@ -120,15 +126,11 @@ class TCSListener(resource.Resource):
         if ("WorkOrderReceipt" in input_json['method']):
             return self.workorder_receipt_handler.workorder_receipt_handler(
                 input_json_str)
-        elif ("Worker" in input_json['method']):
-            return self.worker_registry_handler.worker_registry_handler(
-                input_json_str)
         elif ("EncryptionKey" in input_json['method']):
             return self.worker_encryption_key_handler.process_encryption_key(
                 input_json_str)
         else:
-            logger.info("Received Work Order request : %s",
-                        input_json['method'])
+            logger.info("Received request: %s", input_json['method'])
             # save the full json for WorkOrderSubmit
             input_json["params"]["raw"] = input_json_str
 
--- a/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
+++ b/examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
@@ -14,16 +14,33 @@
 
 import json
 import logging
-import utility.utility as utility
 from error_code.error_status import WorkerError
 from error_code.error_status import WorkerStatus
+
+from jsonrpc.exceptions import JSONRPCDispatchException
 
 logger = logging.getLogger(__name__)
 
 
+def must_get_worker_id(params):
+    if 'workerId' not in params:
+        raise JSONRPCDispatchException(
+            WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE,
+            "Worker Id not found in the database. Hence invalid parameter")
+
+    return str(params['workerId'])
+
+
 class TCSWorkerRegistryHandler:
     """
-     TCSWorkerRegistryHandler processes worker registry requests defined in the TC API spec at Worker Registry Direct API and Worker Type Data API . It reads appropriate worker information from the KV storage to generate the response
+     TCSWorkerRegistryHandler processes worker registry requests defined in the
+     TC API spec at Worker Registry Direct API and Worker Type Data API . It
+     reads appropriate worker information from the KV storage to generate the
+     response.
+     All raised exceptions will be caught and handled by any
+    jsonrpc.dispatcher.Dispatcher delegating work to this handler. In our case,
+    the exact dispatcher will be the one configured by the TCSListener in the
+    ./tcs_listener.py
     """
 # ------------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -236,97 +212,99 @@ def __lookup_basic(self, is_lookup_next, input_value, response):
...
 
-        response["result"] = {
+        result = {
             "totalCount": total_count,
             "lookupTag": lookupTag,
             "ids": ids,
         }
 
-        return response
+        return result
```
manojgop: Should the return value be a well formatted json string object like below ?
{
    "jsonrpc": "2.0",
    "id": <integer>,
    "result": {
        "totalCount":total_count
        "lookupTag":lookupTag,
        "ids": ids,
    }
}

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -163,54 +144,49 @@ def __validate_input_worker_status(self, input_json):
...
-    def __process_worker_set_status(self, worker_id, input_value, response):
+    def WorkerSetStatus(self, **params):
         """
         Function to set the status of worker
         Parameters:
             - worker_id is the worker id specified in the input json str.
-            - input_json_str is a worker request json as per TCF API 5.3.3 Worker Set Status JSON Payload
+            - input_json_str is a worker request json as per TCF API 5.3.3
+                Worker Set Status JSON Payload
             - response is the response object to be returned to client.
```
manojgop: response object in function comment has to be updated with params with this change

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -73,88 +91,51 @@ def __worker_registry_handler_on_boot(self):
...
-                 Hence invalid parameter")
-        return response
+                "Worker Id already exists in the database." +
+                " Hence invalid parameter")
+
+        # Worker Initial Status is set to Active
+        params["status"] = WorkerStatus.ACTIVE
+
+        input_json_str = json.dumps(params)
+        self.kv_helper.set("workers", worker_id, input_json_str)
```
Ram-srini: worker register and worker set status should return functionality not implemented, since there is no functionality in enclave manager side to start the new worker and set the status to other possible values.

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -163,54 +144,49 @@ def __validate_input_worker_status(self, input_json):
...
-            json_dict['status'] = input_value['params']['status']
+            if not self.__validate_input_worker_status(params):
+                raise JSONRPCDispatchException(
+                    WorkerError.INVALID_PARAMETER_FORMAT_OR_VALUE,
+                    "Invalid parameter: worker status code")
+
+            json_dict['status'] = params['status']
             value = json.dumps(json_dict)
             self.kv_helper.set("workers", worker_id, value)
-            response = utility.create_error_response(
```
Ram-srini: It should return functionality not implemented.

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
sammyne: We should just return the `result` part, and the json-rpc library will take care of wrapping the `result` into the formatted JSON-RPC object, which will be returned as `response.data` (a well formatted json string) in L133-L139

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
sammyne: Sorry, I don't got it = =. Would you mind showing me  the reference codes ~ @Ram-srini

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
sammyne: Do you mean replacing the `SUCCESS` with something like `FUNCTIONALITY_NOT_IMPLEMENTED`?

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
Ram-srini: WorkerRegister() and WorkerSetStatus() functionalities are not implemented in the back end(in Enclave manager). For registering new worker should start a new enclave and setting worker status to OFFLINE, COMMISSIONED, etc.. is not happening in the background. Currently only one worker is supported and it is being started by enclave manager during boot time. Time being we should return with new ENUM  FUNCTIONALITY_NOT_IMPLEMENTED until we implement those APIs.

manojgop: @Ram-srini This we should probably handle in another PR. @sammyne has just used existing error codes and error messages as this PR is related to replacing existing implementation with JSON-JRPC library for handling request.

sammyne: Thanks for clarification~ @Ram-srini @manojgop 😄

examples/common/python/connectors/direct/tcs_listener/tcs_listener.py
```diff
@@ -84,6 +84,12 @@ def __init__(self, config):
...
         rpc_methods = [
+            self.worker_registry_handler.WorkerLookUp,
+            self.worker_registry_handler.WorkerLookUpNext,
+            self.worker_registry_handler.WorkerRegister,
+            self.worker_registry_handler.WorkerSetStatus,
+            self.worker_registry_handler.WorkerRetrieve,
+            self.worker_registry_handler.WorkerUpdate,
             self.workorder_handler.WorkOrderSubmit,
             self.workorder_handler.WorkOrderGetResult,
         ]
```
arsulegai: Question: Can there be multiple dispatchers attached. One for registering the workers and another for the work order?

(Probably can be taken up as a next activity to improve the performance)
I see the latter can be large number when considering the scale. If orchestration between `multiple workers for the same task` feature is added later then it could be an advantage. One should be able to add more workers and independent of the work order dispatcher pending queue.

examples/common/python/connectors/direct/tcs_listener/tcs_worker_registry_handler.py
```diff
@@ -163,54 +142,47 @@ def __validate_input_worker_status(self, input_json):
             return False
 
 # ------------------------------------------------------------------------------------------------
-    def __process_worker_set_status(self, worker_id, input_value, response):
+    def WorkerSetStatus(self, **params):
```
Ram-srini: We follow pep 8 coding style, could you please use snake case for method name?

examples/common/python/connectors/direct/tcs_listener/tcs_listener.py
sammyne: I think this can be achieved by implementing those handler as independent listeners, so that every listener can take it own dispatcher ~ @arsulegai 

As for multiple dispatchers within the same listener, I known no way to implement it now = =

arsulegai: Right, let's make a note of this. Let us consider these thoughts in future design considerations/implementations.

PR #134, opened 2019-10-15
Make Heart Disease Worker Consistent with Client

danintel:
* ST depression is a percentage, not a yes/no answer
* Add comments for each risk factor calculation
* Reword result string

### DIFF
--- a/examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
+++ b/examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
@@ -12,9 +12,12 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+
 #include <string>
 #include <cmath>
 #include "heart_disease_evaluation_logic.h"
+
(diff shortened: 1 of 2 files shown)


examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
```diff
@@ -91,27 +100,31 @@ double HeartDiseaseEvalLogic::score_cp(int cp_type) {
...
-        case 0:
+        case 0: // normal
             score = 100.0;
             break;
-        case 1:
+        case 1: // ST-T wave abnormality
             score = 9.0;
             break;
-        case 2:
+        case 2: // hypertrophy
```
manojgop: We need a doctor to review this PR. I don't understand these terms much  :-)

danintel: :-).
Julian Googled many of the variable names so he could name each variable in the GUI form.

danintel:
@arsulegai I forgot the DCO, so I added it.

Issue #19, opened 2019-08-14
default lmdb storage size is too large, causing boot error


m3ngyang:
https://github.com/hyperledger-labs/trusted-compute-framework/blob/fead0d3f715a53cef1079cb271015b27d655bae9/tcs/core/common/packages/db_store/lmdb_store.cpp#L35-L40

The default lmdb storage size is too large, which may cause boot error. Can we set it as a booting config parameter ❓

Ram-srini:
I guess we are using DEFAULT_DB_STORE_SIZE parameter to set the map size using mdb_env_set_mapsize(). It will set the map size to this 1TB, but allocate that much memory affront. As per the mdb API mdb_env_set_mapsize(), size should be a multiple of the OS page size. The default is 10485760 bytes. The size of the memory map is also the maximum size of the database. The value should be chosen as large as possible, to accommodate future growth of the database.

This size can be read from boot config parameter.

danintel:
LMDB files are "sparse" files and do not use disk space until needed. So a 1TB sparse file does not use 1TB of disk storage.

To copy or backup these files, Linux commands have sparce file options, such as cp --sparse or tar -S

m3ngyang:
closed by #45

PR #137, opened 2019-10-16
Resolve container caching issue in Jenkins

pankajgoyal2:
Each Jenkins job creates a container for TCF build and tests.
If container_name is given in compose file, each jenkins job
will create and destroy the container with name "tcf", which is
not a good situation for parallel jenkins jobs. If, container_name
is not given in compose files, containers will automatically be
assigned with a unique name using COMPOSE_PROJECT_NAME variable
from Jenkinsfile. So, removing container_name from compose files.

### DIFF
--- a/ci/docker-compose-build.yaml
+++ b/ci/docker-compose-build.yaml
@@ -16,7 +16,6 @@ version: '3.5'
 
 services:
   tcf:
-    container_name: tcf
     image: tcf-ci-build:${ISOLATION_ID}
 
     build:
--- a/ci/docker-compose-tests.yaml
+++ b/ci/docker-compose-tests.yaml
@@ -16,7 +16,6 @@ version: '3.5'
 
 services:
   tcf:
-    container_name: tcf
     image: tcf-ci-build:${ISOLATION_ID}
     build:
       context: .


pankajgoyal2:
> The problem with removing the container name is it makes TCF harder to use with docker-compose. Now, to start a bash shell, one types: `sudo docker exec -it tcf bash`. Removing the container name will break the instructions in `BUILD.md`
> 
> Maybe there can be a different docker-compose file just for use with Jenkins without a `container_name`.
> 
> Or maybe the Jenkins job can delete the cached containers. Not having `tcf` as a container name does not delete the old cached containers. Do they stay around forever?
> 
> I notice the `sawtooth-core` project Docker files have `container_name` and they use Jenkins OK.

1. The changes that I have done are in compose files in "ci" folder. I guess these compose files are not part of documentation in .BUILD.md.
2. Containers are being killed at the end of compose command. The problem occurs when we trigger 2 simultaneous jenkins builds. For eg: in http://13.232.103.220:8080/job/trusted-compute-framework/job/master/ I started Build #19 first and then Build #20. Build #20 is failing because of this issue
---------------------
ERROR: for tcf  Cannot create container for service tcf: Conflict. The container name "/tcf" is already in use by container "c19f48d374c8b3f4b3a69349fc3a56d1325afd0e46cd010d61f4b0cc4cece141". You have to remove (or rename) that container to be able to reuse that name.
---------------------
3. I see the sawtooth-core is also using container_name but it is just building the image. Creating the container out of this image is done seperately through a shell script while running the tests. But in TCF, it is mandatory to call "docker-compose up" even to build the TCF because of entrypoint part in compose file that builds the C++ code.

May be there is another better solution to it. Any suggestions are welcome from everyone.

arsulegai:
> The problem with removing the container name is it makes TCF harder to use with docker-compose. Now, to start a bash shell, one types: `sudo docker exec -it tcf bash`. Removing the container name will break the instructions in `BUILD.md`
> 
> Maybe there can be a different docker-compose file just for use with Jenkins without a `container_name`.
> 
> Or maybe the Jenkins job can delete the cached containers. Not having `tcf` as a container name does not delete the old cached containers. Do they stay around forever?
> 
> I notice the `sawtooth-core` project Docker files have `container_name` and they use Jenkins OK.

I agree with @danintel on this. The build compose can be different from the one we use to run if it's done that way. The example `sawtooth-core` does that, there are separate files for building and running the code. The `*-build` can be kept for the build only purpose. @pankajgoyal2 Could you please confirm it's not the case here?

pankajgoyal2:
> > The problem with removing the container name is it makes TCF harder to use with docker-compose. Now, to start a bash shell, one types: `sudo docker exec -it tcf bash`. Removing the container name will break the instructions in `BUILD.md`
> > Maybe there can be a different docker-compose file just for use with Jenkins without a `container_name`.
> > Or maybe the Jenkins job can delete the cached containers. Not having `tcf` as a container name does not delete the old cached containers. Do they stay around forever?
> > I notice the `sawtooth-core` project Docker files have `container_name` and they use Jenkins OK.
> 
> I agree with @danintel on this. The build compose can be different from the one we use to run if it's done that way. The example `sawtooth-core` does that, there are separate files for building and running the code. The `*-build` can be kept for the build only purpose. @pankajgoyal2 Could you please confirm it's not the case here?

@arsulegai Please read my reply to @danintel

arsulegai:
> > > The problem with removing the container name is it makes TCF harder to use with docker-compose. Now, to start a bash shell, one types: `sudo docker exec -it tcf bash`. Removing the container name will break the instructions in `BUILD.md`
> > > Maybe there can be a different docker-compose file just for use with Jenkins without a `container_name`.
> > > Or maybe the Jenkins job can delete the cached containers. Not having `tcf` as a container name does not delete the old cached containers. Do they stay around forever?
> > > I notice the `sawtooth-core` project Docker files have `container_name` and they use Jenkins OK.
> > 
> > 
> > I agree with @danintel on this. The build compose can be different from the one we use to run if it's done that way. The example `sawtooth-core` does that, there are separate files for building and running the code. The `*-build` can be kept for the build only purpose. @pankajgoyal2 Could you please confirm it's not the case here?
> 
> @arsulegai Please read my reply to @danintel

Oops! I missed it. Yes, the first bullet makes it clear. Thanks :)

danintel:
I approve. I thought this was changing `docker-compose.yaml`, but it changes `ci/docker-compose-build.yaml`

PR #140, opened 2019-10-18
Heart Disease GUI Improvements

danintel:
* Add option to generate random input data
* Check for empty input number string
* Split Logger and TCF initialization out of main method
* Make GUI screen more narrow
* Add Python Docstrings
* Wrap lines > 80 characters


danintel:
Fixed alerts from LGTM (duplicate import and change `== None` to `is None`)

PR #142, opened 2019-10-21
Remove virtual environment references in docker

manju956:
The reason for having virtual environment is to provide isolation from TCF execution environment from the host system. Since, docker provides isolation from the host, we don't need virtual environment inside docker environment and hence the references are removed.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -53,16 +53,8 @@ Follow the instructions below to execute a Docker-based build and execution.
    followed by a repetitive message `Enclave manager sleeping for 10 secs`
 4. Open a Docker container shell using following command
    `sudo docker exec -it tcf bash`
-5. Activate a Python virtual environment in the Docker shell using following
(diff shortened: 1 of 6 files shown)


manju956:
> environment?

Hi Dan,
The reason for having virtual environment is to provide isolation for the TCF dependencies from the host system. Since, Docker already provides isolation from the host, we don't need virtual environment inside Docker. 
All downloaded packages and TCF python module packages are installed in the standard library path /usr/lib/python3/dist-packages. 

I will also update the reason for removal of virtual environment in the PR message.

arsulegai:
Ideally, we should not have separate installation instructions for the Docker. Maybe the next step is to remove any such references as well.

danintel:
This PR has conflicts with BUILD.md that need to be resolved before this PR can be pulled into master.

manju956:
> This PR has conflicts with BUILD.md that need to be resolved before this PR can be pulled into master.

Rebase done.

PR #143, opened 2019-10-22
Add Accumulated Heart Disease History to GUI and Worker

danintel:
- Add ability to return accumulated heart disease history in worker and client
- Change TCF to Avalon
- Update screenshot, heart_gui_screenshot.jpg


danintel:
Here is an example showing the accumulated results in the GUI:
heart-disease-demo-accumulated-results-small

examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
```diff
@@ -223,17 +222,11 @@ double HeartDiseaseEvalLogic::score_num(int num) {
...
         std::string decrypted_user_input_str) {
+    std::string resultString;
     // Variables to accumulate multiple results
     static int totalRisk = 0;
     static int count = 0;
 
-    // Clear accumulation data if input is empty
-    if (decrypted_user_input_str.empty()) {
-        totalRisk = count = 0;
-        return "";
```
manojgop: How are we resetting the totalRisk and count with new logic ? Or is it not required since every new work order request will result in creation of new instance of  work order processor ?

examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
```diff
@@ -242,31 +235,38 @@ std::string HeartDiseaseEvalLogic::executeWorkOrder(
...
-            + score_thaldur(std::stoi(medData[12])) * 0.02
-            + score_num(std::stoi(medData[13])) * 0.04 );
-
-        // Update accumulation
-        totalRisk += risk;
-        count++;
-        // Use accumulated data to calculate the result
-        resultString = "You have a " +
-            std::to_string(totalRisk/count) + "% risk of heart disease";
+        if (medData.size() != 14) { // return accumulated calculations
```
manojgop: Earlier this check was used to validate the input format. With this change if input format is incorrect ( not passing required number of input values), we are not returning error message. What is the expected way to use Heart disease workload with command line (eg: with generic client).

examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
danintel: I now use `risk` instead of `totalRisk` to calculate the risk of heart disease for one person (line 267).
The original intent of `totalRisk` is for it to be a total of all the risk percentages for all people.  Dividing that by the count gives an average risk: `totalRisk / count`  (line 243).

No, each work order request does not create a new worker instance. That is why I can calculate an average risk (in addition to an individual risk)

examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
danintel: When the worker receives 0 data items it returns an average risk.
When it receives 14 data items it returns the individual risk.
I will change the code to check for either 0 or 14 data items and return an error for other counts.

examples/apps/heart_disease_eval/workload/heart_disease_evaluation_logic.cpp
danintel: @manojgop This works good enough for demo purposes using one worker.
But I will add IPC so multiple workers can be used.

manojgop: @danintel . we have only single worker (one enclave) with multiple workload (heart-disease, echo etc). And currently the worker implementation is stateless. For every new work order request , a new instance of workloadProcessor object instance is created.  The scope of static variables (totalRisk,count) is valid only till workloadProcessor object is valid. So even for heart disease eval workload, to calculate aggregate we need pass all input data in single work order request.We can't aggregate the results over multiple work order requests.

manojgop: Looks like value of static variable inside class function is persevered across different instances of class.  I initial thought we have to declare it as static member variable of class to preserve the value across multiple instances.

danintel: @manojgop The static variable is only shared for a single process. So more work needs to be done to share it across processes--with multiple instances.

PR #146, opened 2019-10-24
Fix Generic CLI Client Error Handling and Mode Default

danintel:
- Add default to `-m`/`--mode` command line option
- Change `logger.info` or `logger.warn` to `logger.error` if before `sys.exit`

### DIFF
--- a/examples/apps/generic_client/generic_client.py
+++ b/examples/apps/generic_client/generic_client.py
@@ -66,7 +66,9 @@ def ParseCommandLine(args) :
 		help="an address (hex string) of the smart contract (e.g. Worker registry listing)",
 		type=str)
 	parser.add_argument("-m", "--mode",
-		help="should be one of listing or registry",
+		help="should be one of listing or registry (default)",
+		default="registry",		
+		choices={"registry", "listing"},
 		type=str)
 	parser.add_argument("-w", "--worker_id",
 		help="worker id (hex string) to use to submit a work order",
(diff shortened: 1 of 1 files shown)


danintel:
A symptom of the bug is this error message
`Address needs to be passed with mode option`

examples/apps/generic_client/generic_client.py
```diff
@@ -66,7 +66,8 @@ def ParseCommandLine(args) :
 		help="an address (hex string) of the smart contract (e.g. Worker registry listing)",
 		type=str)
 	parser.add_argument("-m", "--mode",
-		help="should be one of listing or registry",
+		help="should be one of listing or registry (default)",
```
arsulegai: I found an answer on StackOverflow, can that be used here to reduce boilerplate code?
https://stackoverflow.com/questions/15836713/allowing-specific-values-for-an-argparse-argument

danintel: Thanks.I added `               choices={"registry", "listing"}, for `--mode`

examples/apps/generic_client/generic_client.py
```diff
@@ -119,16 +121,18 @@ def ParseCommandLine(args) :
 			sys.exit(-1)
 		else:
 			logger.error("Mode should be either registry or listing")
+			sys.exit(-1)
 
 	else:
 		if not mode:
 			logger.error("Address needs to be passed with mode option")
+			sys.exit(-1)
```
manojgop: Will the mode be set to default "registry" if not explicitly set with the above changes. If so, mode can never be None. Hence "if not mode" condition check is not required I think.

danintel: I removed the code (lines 122-129 above, both `else` clauses)

danintel: Correct.  I removed the `if not mode` check.

PR #148, opened 2019-10-28
Build TCF code as part of docker image

manju956:
TCF code is built outside docker image. this leads to missing TCF packages or
libraries when build and bringing up TCF components/test case execution
does not happen in single docker process (this approach is being followed in CI).
Hence move building TCF components within docker image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,10 +1,8 @@
 #Ignore Kv files
 config/Kv*
 
-#Ignore build directories
-workloads/build*
-tcs/core/common/build*
-tcs/core/tcs_trusted_worker_manager/enclave/build*
-tcs/core/common/python/build*
-common/build*
+#Ignore documentation files
+*.md
+docs/*
+CODEOWNERS
 
--- a/ci/docker-compose-build.yaml
+++ b/ci/docker-compose-build.yaml
@@ -19,16 +19,9 @@ services:
     image: tcf-ci-build:${ISOLATION_ID}
 
     build:
-      context: .
-      dockerfile: ../docker/Dockerfile.tcf-dev
+      context: ..
+      dockerfile: docker/Dockerfile.tcf-dev
       args:
         - http_proxy
         - https_proxy
         - no_proxy
-    volumes:
-      - ../:/project/TrustedComputeFramework
-    working_dir: "/project/TrustedComputeFramework/tools/build"
-    entrypoint: "bash -c \"\
-        if [ ${MAKECLEAN:-1} != 0 ]; then echo \\\"Clean build\\\" && make clean; fi \
-        && make \
-        && echo \\\"BUILD SUCCESS\\\" \""
--- a/ci/docker-compose-tests.yaml
+++ b/ci/docker-compose-tests.yaml
@@ -23,11 +23,8 @@ services:
         - http_proxy
         - https_proxy
         - no_proxy
-    volumes:
-      - ../:/project/TrustedComputeFramework
-    working_dir: "/project/TrustedComputeFramework/tools/build"
+    working_dir: "/project/TrustedComputeFramework/"
     entrypoint: "bash -c \"\
-        cd ../../scripts \
-        && source tcs_startup.sh -y \
+        source scripts/tcs_startup.sh -y \
         && sleep 5 \
-        && source ../tools/run_tests.sh \""
+        && source tools/run_tests.sh \""
--- a/docker-compose-sgx.yaml
+++ b/docker-compose-sgx.yaml
@@ -25,22 +25,18 @@ services:
         - http_proxy
         - https_proxy
         - no_proxy
-    environment:
+        - SGX_MODE=HW
         - TCF_DEBUG_BUILD=${TCF_DEBUG_BUILD:-}
         - DISPLAY=${DISPLAY:-}
+        - MAKECLEAN=${MAKECLEAN:-1}
     volumes:
-      - ./:/project/TrustedComputeFramework
+      # Below volume mappings are required for DISPLAY settings by heart disease GUI client
       - /tmp/.X11-unix:/tmp/.X11-unix
+      - ~/.Xauthority:/root/.Xauthority
     devices:
       - "/dev/isgx:/dev/isgx"
-    working_dir: "/project/TrustedComputeFramework/tools/build"
-    environment:
-      - SGX_MODE=HW
+    working_dir: "/project/TrustedComputeFramework"
     entrypoint: "bash -c \"\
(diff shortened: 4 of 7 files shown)


ci/docker-compose-build.yaml
```diff
@@ -19,16 +19,11 @@ services:
...
-      context: .
-      dockerfile: ../docker/Dockerfile.tcf-dev
+      context: ..
+      dockerfile: docker/Dockerfile.tcf-dev
       args:
         - http_proxy
         - https_proxy
         - no_proxy
     volumes:
       - ../:/project/TrustedComputeFramework
```
arsulegai: Is volume mounting required after this change?

ci/docker-compose-tests.yaml
```diff
@@ -25,9 +25,8 @@ services:
         - no_proxy
     volumes:
```
arsulegai: Question: Will this work if the host machine doesn't have the build available? The TCF directory is replaced with the host machine files. I have similar questions for the other files in this commit.

arsulegai:
I am unable to block the commit, once this is moved out from the hyperledger-labs repository. Request you to please address the concerns around the volume mounting.

manju956:
> I am unable to block the commit, once this is moved out from the hyperledger-labs repository. Request you to please address the concerns around the volume mounting.

I will address volume mapping concerns and update the PR

ci/docker-compose-build.yaml
manju956: Not needed, will remove it

ci/docker-compose-tests.yaml
manju956: Since TCF build is part of docker image with these changes, test case execution will succeed because all build artifacts are already part of docker image.

manju956: But i have removed volume mapping from the compose file, hence libraries built as part of docker image will be used.

docker-compose-sgx.yaml
```diff
@@ -25,17 +25,15 @@ services:
...
         - https_proxy
         - no_proxy
-    environment:
+        - SGX_MODE=HW
         - TCF_DEBUG_BUILD=${TCF_DEBUG_BUILD:-}
         - DISPLAY=${DISPLAY:-}
+        - MAKECLEAN=${MAKECLEAN:-1}
     volumes:
-      - ./:/project/TrustedComputeFramework
       - /tmp/.X11-unix:/tmp/.X11-unix
```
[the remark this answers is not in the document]
manju956: Ok

ci/docker-compose-tests.yaml
arsulegai: Right! After the change it won't map.

docker-compose-sgx.yaml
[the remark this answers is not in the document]
arsulegai: Thanks.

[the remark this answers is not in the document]
danintel: The X11 directory mapping is required enable to allow running X applications (such as the heart disease GUI).

.dockerignore
```diff
@@ -2,9 +2,10 @@
...
-tcs/core/common/build*
-tcs/core/tcs_trusted_worker_manager/enclave/build*
-tcs/core/common/python/build*
-common/build*
+tc/sgx/trusted_worker_manager/enclave/build*
+common/sgx_workload/build*
+tc/sgx/common/build*
+examples/enclave_manager/build*
+examples/common/python/build*
+examples/apps/build*
```
manojgop: Will the build directories still created in host machine or are build artifacts created only in docker image and visible only to the container ? If build directories are visible only to container then can we remove these lines ?

ci/docker-compose-tests.yaml
```diff
@@ -23,11 +23,8 @@ services:
         - http_proxy
         - https_proxy
         - no_proxy
-    volumes:
-      - ../:/project/TrustedComputeFramework
-    working_dir: "/project/TrustedComputeFramework/tools/build"
+    working_dir: "/project/TrustedComputeFramework/"
```
manojgop: WORKDIR is set in Dockerfile. Do we still need to set this in docker-compose file. Since the build is done from Dockerfile, Can we remove this line in docker-compose file ? I notices that "working_dir" is different for ci/docker-compose-test.yaml and docker-compose.yaml. Is that intentional ?

docker-compose-sgx.yaml
```diff
@@ -25,17 +25,17 @@ services:
...
+      - ~/.Xauthority:/root/.Xauthority
     devices:
       - "/dev/isgx:/dev/isgx"
     working_dir: "/project/TrustedComputeFramework/tools/build"
-    environment:
-      - SGX_MODE=HW
     entrypoint: "bash -c \"\
         if [ ${MAKECLEAN:-1} != 0 ]; then echo \\\"Clean build\\\" && make clean; fi \
         && make \
```
manojgop: This needs to be removed since its part of Dockerfile

docker-compose-sgx.yaml
```diff
@@ -25,17 +25,17 @@ services:
...
         - DISPLAY=${DISPLAY:-}
+        - MAKECLEAN=${MAKECLEAN:-1}
     volumes:
-      - ./:/project/TrustedComputeFramework
+      # Below volume mappings are required for DISPLAY settings by heart disease GUI client
       - /tmp/.X11-unix:/tmp/.X11-unix
+      - ~/.Xauthority:/root/.Xauthority
     devices:
       - "/dev/isgx:/dev/isgx"
     working_dir: "/project/TrustedComputeFramework/tools/build"
```
manojgop: working_dir may not be needed here as build is done inside Dockerfile ?

.dockerignore
manju956: Build artifacts will not be created on the host machine and are only created in the docker image.
I will remove build directories from dockerignore file.

docker-compose-sgx.yaml
manju956: Done

.dockerignore
```diff
@@ -1,10 +1,13 @@
 #Ignore Kv files
 config/Kv*
 
-#Ignore build directories
-workloads/build*
-tcs/core/common/build*
-tcs/core/tcs_trusted_worker_manager/enclave/build*
-tcs/core/common/python/build*
-common/build*
+#Ignore documentation files
```
manojgop: You can use *.md  and docs/\* to exclude all documentation files.

manju956: Done

PR #160, opened 2019-11-04
Add Intel SGX Detection and Enablement

danintel:
- Add how to detect Intel SGX supported
- Mention Intel SGX must be enabled in BIOS

### DIFF
--- a/PREREQUISITES.md
+++ b/PREREQUISITES.md
@@ -174,11 +174,21 @@ install the Intel SGX driver for both standalone and docker builds.
 You need to install the Intel SGX driver whether you build Avalon standalone
 or using Docker.
 
-First install this package:
-
+Before installing Intel SGX software, install this package:
 ```
-sudo apt-get install -y libelf-dev
+sudo apt-get install -y libelf-dev cpuid
 ````
+
+Verify your processor supports Intel SGX with:
+`cpuid | grep SGX:`
+
+Verify Intel SGX is enabled in BIOS.
+Enter BIOS by pressing the BIOS key during boot.
+The BIOS key varies by manufacturer and could be F10, F2, F12, F1, DEL, or ESC.
+Usually Intel SGX is disabled by default.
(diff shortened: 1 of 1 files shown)


PREREQUISITES.md
```diff
@@ -174,8 +174,10 @@ install the Intel SGX driver for both standalone and docker builds.
 You need to install the Intel SGX driver whether you build Avalon standalone
 or using Docker.
 
-First install this package:
+Verify your processor supports Intel SGX with `cpuid | grep SGX:` .
+Also verify Intel SGX is enabled in BIOS.
```
manojgop: How about giving the link to this video https://www.youtube.com/watch?v=bca5NcjoEdc

PREREQUISITES.md
```diff
@@ -418,6 +422,9 @@ problems.
   you are using an old version of OpenSSL and need to clone the
   `openssl_1.1.1` branch of `intel-sgx-ssl` in the step above
 
+- If the message  `intel_sgx: SGX is not enabled` appears in `/var/log/syslog`
+  Intel SGX needs to be enabled in BIOS
```
manojgop: Is this a standard recommended way to detect and enable SGX ?

danintel: No. It is only how to detect in Linux if SGX is enabled in BIOS.

This message appears even if SGX is supported by the processor an even if drivers are installed.

PREREQUISITES.md
danintel: I like John's SGX videos but I do not want to require watching a video to install SGX or Avalon.

manojgop: What I meant is it's not clear how to verify if Intel SGX is enabled in BIOS ? or is this not part of documentation scope ?

dcmiddle: cpuid may not be installed on all distros. For example, I don't think it's part of the ubuntu docker image Avalon uses. 
`cat /proc/cpuinfo | grep sgx` 
might be more portable.

PR #161, opened 2019-11-05
Add vim to dev container

dcmiddle:
Add a basic editor to the dev container. Facilitates basic debug and
editing.

Signed-off-by: Dan Middleton <dan.middleton@intel.com>

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -68,6 +68,7 @@ RUN apt-get update \
     liblmdb-dev \
     ocamlbuild \
     python3-tk \
+    vim \
  && apt-get -y -q upgrade \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*


docker/Dockerfile.tcf-dev
```diff
@@ -68,6 +68,7 @@ RUN apt-get update \
     liblmdb-dev \
     ocamlbuild \
     python3-tk \
+    vim \
```
danintel: I don't like the idea of adding development tools to the Docker container.  I don't see that being done with any other Dockerfile I looked at. Vim and other tools can always be added later.

If you must add `vim`, please add `vim-tiny` instead of the full-blown `vim` family of packages and dependencies.

jopasserat:
Why not editing with a Vim (or other) installed on your own dev box since all the files of the project are available in and outside the container? :thinking:

docker/Dockerfile.tcf-dev
EugeneYYY: I agree with previous comment that we should minimize number of components in the container.
VIM (or other editor) can be helpful for debugging and it can be achieved by
- including an editor to the container as needed during the development
- editing files on the host since all of them available
- creating a different container (debugger version). This debugger container maybe included in the repo with an appropriate documentation explaining how to use it instead of primary (minimally sized) container

PR #166, opened 2019-11-07
Unit-tests for Utility modules

rranjan3:
1. Add unit-tests for file/hex/utility modules
2. Bug fixes in these modules
3. Refactor functions
4. Include nose2 and invocation in ci build


arsulegai:
Could you please create 4 commits for the 4 points you mentioned?

examples/common/python/utility/file_utils.py
```diff
@@ -55,7 +60,32 @@ def read_json_file(input_file, data_dir = ['./', '../', '/']) :
...
+#---------------------------------------------------------------------------------------------
+def write_result_data_to_json_file(file_name, input_data, data_dir ='./') :
+    """
+    Function to store result data as json file
+    Parameters:
+        - file_name is the name in which the file should be stored
+        - input_data is a JSON data which needs to be stored in a file( should have attribute 'result')
+        - data_dir is the directory path to store the file
+    """
+    result_info = dict()
```
pankajgoyal2: I think, the preferable way to declare an empty dictionary in python is result_info = {}. This way is faster.
Your way of declaring empty dictionary is good as it provides flexibility in syntax and initialization. But  as you are declaring an empty dict, I recommend to use {}.

examples/common/python/config/config.py
```diff
@@ -27,9 +27,14 @@
...
-from utility.file_utils import find_file_in_path
+from utility.file_utils import find_file_in_paths
 
-__all__ = [ "ConfigurationException", "parse_configuration_files", "parse_configuration_file" ]
+__all__ = [ "ConfigurationException", 
+    "parse_configuration_files", 
+    "parse_configuration_file", 
+    "read_config_from_toml" ]
+
+TCFHOME = os.environ.get("TCF_HOME", "../../")
```
pankajgoyal2: for TCFHOME, the default is set to  2 directories up (../../) i.e. examples/common folder. Doesn't it has to be 4 levels up (../../../../)?

rranjan3:
> Could you please create 4 commits for the 4 points you mentioned?

Done. Broken down to 3 commits.

examples/common/python/config/config.py
```diff
@@ -104,3 +109,27 @@ def parse_configuration_file(filename, variable_map) :
...
+# -----------------------------------------------------------------
+# -----------------------------------------------------------------
+
+def read_config_from_toml(input_file, config_name = None, confpaths = [".", TCFHOME + "/" + "config"]):
+    """
+    Function to read toml file and returns the toml content as a list
+    Parameters:
+        - input_file is any toml file which need to be read
+        - config_name is particular configuration to pull
+        - data_dir is the directory structure in which the toml file exists
```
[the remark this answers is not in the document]
rranjan3: Updated.

tests/test_utility/test_file_utils/test.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2019 Intel Corporation
```
theramkrish8: Can we add these unit tests in the respective module directory by creating folder unit_tests that module? If future if move these module to some other directory then we can module as it is.

examples/common/python/config/config.py
```diff
@@ -104,3 +109,27 @@ def parse_configuration_file(filename, variable_map) :
...
+    """
+    conf_files = [input_file]
+    config = parse_configuration_files(conf_files, confpaths)
+    if config_name is None:
+        return config
+    else :
+        result = config.get(config_name)
+        if result is None:
+            logger.error("%s is missing in toml file %s", config_name, input_file )
+            return None
```
manojgop: We can `return result` as its already `None` in this case.  we can have a common `return result` for if/else case
```
if result is None
    logger.error("%s is missing in toml file %s", config_name, input_file )
return result
```

tests/test_utility/test_file_utils/test.py
rranjan3: That's what I started with. But later found out that there was a tests directory at the root, so decided to rather put it along side the other tests. I will raise another PR to refactor a few thing, might do it there. There is a lot of scope for improvement. For example we could make the tests modular enough to be run with choice or all at once. Needs some thorough thinking.

examples/common/python/config/config.py
rranjan3: Thanks Manoj. I had only moved this function not looking into the implementation. Fixed.

examples/common/python/config/config.py
rranjan3: Good catch Pankaj. I moved the statement(along with a function)  from another module. Though I have removed the default option altogether now from the os.environ.get method. That argument needs to be checked throughout. Relative paths can be misleading depending on the current working directory.

examples/common/python/utility/file_utils.py
rranjan3: This again is a moved piece of code Pankaj. However I would rather prefer it to be this way(more verbose the better). Performance might not be an issue here at all as this is a standalone statement(not within a loop as such) and called rather infrequently. By the way I did evaluate the efficiency of the two as well -
```
C:\Users\rranjan3>python -m timeit -n 1000000 -r 5 -v 'dict()'
raw times: 12.4 msec, 8.67 msec, 9.71 msec, 8.77 msec, 8.42 msec

1000000 loops, best of 5: 8.42 nsec per loop

C:\Users\rranjan3>python -m timeit -n 1000000 -r 5 -v '{}'
raw times: 16.2 msec, 8.5 msec, 9.92 msec, 9.88 msec, 9.51 msec

1000000 loops, best of 5: 8.5 nsec per loop
```

They seem to be very much at par.

examples/common/python/config/config.py
```diff
@@ -27,9 +27,14 @@
...
-from utility.file_utils import find_file_in_path
+from utility.file_utils import find_file_in_paths
 
-__all__ = [ "ConfigurationException", "parse_configuration_files", "parse_configuration_file" ]
+__all__ = [ "ConfigurationException", 
+    "parse_configuration_files",
+    "parse_configuration_file",
+    "read_config_from_toml" ]
+
+TCFHOME = os.environ.get("TCF_HOME")
```
pankajgoyal2: As you have removed the default relative path, now you should check if TCF_HOME is defined as environment variable or not. If not, then raise exception.
try:  
   os.environ["TCF_HOME"]
except KeyError: 
   print "Please set the environment variable TCF_HOME"
   sys.exit(1)

examples/common/python/utility/file_utils.py
pankajgoyal2: makes sense and good exploration!

examples/common/python/config/config.py
```diff
@@ -104,3 +109,25 @@ def parse_configuration_file(filename, variable_map) :
         text = Template(text).substitute(variable_map)
 
     return toml.loads(text)
+
+# -----------------------------------------------------------------
+# -----------------------------------------------------------------
+
+def read_config_from_toml(input_file, config_name = None, confpaths = [".", TCFHOME + "/" + "config"]):
```
pankajgoyal2: remove space around = for function arguments with default values

examples/common/python/utility/file_utils.py
```diff
@@ -55,7 +60,31 @@ def read_json_file(input_file, data_dir = ['./', '../', '/']) :
...
-        input_json = input_json_file.read().rstrip('\n')
-    return input_json
+        input_lines = input_json_file.read().splitlines()
+    flattened_json = ""
+    for line in input_lines :
+        flattened_json += line
+    return flattened_json
+
+#---------------------------------------------------------------------------------------------
+def write_result_data_to_json_file(file_name, input_data, data_dir ='./') :
```
pankajgoyal2: 1. You have introduced space before ":" above. There is an inconsistency in the code with this. The preferred way is not to introduce space before ":". This inconsistency is not because of your code. It is already there.
2. PEP-8 says, don't introduce space around = for function arguments with default values. So, it should be data_dir='./'

examples/common/python/config/config.py
rranjan3: Updated this to handle the error case. Rather than using exit, raising a KeyError.

examples/common/python/utility/file_utils.py
```diff
@@ -55,7 +60,31 @@ def read_json_file(input_file, data_dir = ['./', '../', '/']) :
...
     """
-    file_name = find_file_in_path(input_file, data_dir)
+    file_name = find_file_in_paths(input_file, data_dir)
     with open(file_name, "r") as input_json_file :
-        input_json = input_json_file.read().rstrip('\n')
-    return input_json
+        input_lines = input_json_file.read().splitlines()
+    flattened_json = ""
+    for line in input_lines :
+        flattened_json += line
```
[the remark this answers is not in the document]
rranjan3: Updated to remove an extra variable there but not same as you suggested. Just to ensure readability.

examples/common/python/config/config.py
rranjan3: Done.

examples/common/python/utility/file_utils.py
rranjan3: Updated the 2nd one. Left the first one as is. We can integrate the PEP8 tool to catch such issues and carry out a comprehensive cleanup in that case as a single task.

tests/test_utility/test_utility.py
```diff
@@ -0,0 +1,96 @@
...
+        expected_response = {"jsonrpc":"2.0","id":"id2","error":{"code":"2","message":"General error"},}
+        self.assertEquals(expected_response, create_error_response("2", "id2","General error"))
+
+    def test_strip_begin_end_key(self):
+        """Tests to verify strip_begin_end_key(key) function
+        """
+        expected = "123456aghdfgasdasdkommkf"       # Positive case
+        self.assertEquals(expected, strip_begin_end_key("-----BEGIN PUBLIC KEY-----123456aghdfgas\ndasdkommkf-----END PUBLIC KEY-----"))
+
+        expected = "123456aghdfgasdasdkommkf"       # Interspeced with '\n'
```
danintel: s/Interspeced/Interspersed/

examples/common/python/utility/file_utils.py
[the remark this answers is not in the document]
pankajgoyal2: cool!

tests/test_utility/test_utility.py
rranjan3: Thanks Dan. Updated. I wonder how I typed that. :)

PR #171, opened 2019-11-18
Local docker image for kubernetes

manojgop:
Use locally built docker image for kubernetes setup

### DIFF
--- a/docker/k8s/README.md
+++ b/docker/k8s/README.md
@@ -14,12 +14,16 @@ in SGX-SIM mode.
     ```bash
     kubectl create -f tcf-config-map.yaml
     ```
-2. Start the LMDB server
+2. Enable `remote_url` in [tcs_config.toml](../../config/tcs_config.toml)
+   set `remote_url` to lmdb service defined in lmdb.yaml.
+   eg: `remote_url="http://lmdb:9090"`
+
+3. Start the LMDB server.
     ```bash
     # bootstrap the deployment and service
     kubectl create -f lmdb.yaml
     ```
-3. Start the `EnclaveManager` (to manage SGX enclaves) and `TCSServer` (to 
+4. Start the `EnclaveManager` (to manage SGX enclaves) and `TCS Listener` (to
    accept and handle requests from clients)  
     ```bash
     kubectl create -f enclave-manager-deployment.yaml
--- a/docker/k8s/enclave-manager-deployment.yaml
+++ b/docker/k8s/enclave-manager-deployment.yaml
@@ -14,8 +14,9 @@ spec:
         app: tee
     spec:
       containers:
-      - image: sammyne/hyperledger-tcf
+      - image: tcf-dev:latest
         name: tee
+        imagePullPolicy: Never
         envFrom:
         - prefix: TCF_
           configMapRef:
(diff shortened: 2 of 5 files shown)


docker/k8s/README.md
```diff
@@ -14,12 +14,16 @@ in SGX-SIM mode.
     ```bash
     kubectl create -f tcf-config-map.yaml
     ```
-2. Start the LMDB server
+2. Enable `remote_url` in [tcs_config.toml](../../config/tcs_config.toml)
```
manju956: Instead of relative path, can we give full path - TCFHOME/config/tcs_config.toml?

manojgop: This is a hyperlink which anyone can click and traverse.

docker/k8s/README.md
```diff
@@ -14,12 +14,16 @@ in SGX-SIM mode.
...
+   set `remote_url` to lmdb service defined in lmdb.yaml.
+   eg: `remote_url="http://lmdb:9090"`
+
+3. Start the LMDB server.
     ```bash
     # bootstrap the deployment and service
     kubectl create -f lmdb.yaml
     ```
-3. Start the `EnclaveManager` (to manage SGX enclaves) and `TCSServer` (to 
+4. Start the `EnclaveManager` (to manage SGX enclaves) and `TCSServer` (to
```
Ram-srini: What is TCSServer component is about? Is it listener?

docker/k8s/enclave-manager-deployment.yaml
```diff
@@ -14,8 +14,9 @@ spec:
         app: tee
     spec:
       containers:
-      - image: sammyne/hyperledger-tcf
+      - image: tcf-dev:latest
         name: tee
+        imagePullPolicy: Never
```
Ram-srini: Do we need to set image pull policy to Never always or is it a temporary approach for now?

docker/k8s/README.md
manojgop: Yeah. Looks like original author meant TCS Server is TCS Listener

docker/k8s/enclave-manager-deployment.yaml
manojgop: It is needed now so that Kubernetes doesn't attempt to pull the image since we are using local built.  Later on when we plan to host the Avalon docker image in a external repository, imagePullPolicy need not be set to never

docker/k8s/propose-requests-job.yaml
```diff
@@ -11,7 +11,8 @@ spec:
       restartPolicy: OnFailure
       containers:
       - name: request-proposer
-        image: sammyne/hyperledger-tcf
+        image: tcf-dev
```
pankajgoyal2: did you miss the TAG here?

manojgop: Yeah. I will explicitly add latest build tag

manojgop: Done

docker/k8s/README.md
manju956: Ok

manojgop:
@sammyne Could you review this PR

docker/k8s/enclave-manager-deployment.yaml
arsulegai: Other way to handle this would be to change the tag from `:latest` to either `:local` or `:<BUILD_VERSION>`. The `:<BUILD_VERSION>` is auto generated for every build, this is preferable. Kubernetes tries to pull the images if the tag is `:latest`.

PR #173, opened 2019-11-20
Reorganize work order interface files

manju956:
All sub modules in core common i.e tc/sgx/common are intended to be used
by both trusted and untrusted code. Since work order interface file is used
only in trusted code(shared between enclave and workloads),
they are moved to common/sgx_workload.

Also, creation of seperate static library for work order interface is avoided
and they will be available in common/sgx_workload static library only.


manju956:
> trusted code (shared between enclave and workloads),

Thanks for the suggestion. I will add README file in common/sgx folder.

PR #175, opened 2019-11-23
Enable Checking with pycodestyle for examples/enclave_manager

danintel:
This is phase 2 of pycodestyle changes.


manojgop:
@danintel Are these any tools available to make changes to files for pycodestyle or do we need to make the changes  manually ?

danintel:
@manojgop I know of no tools. That would be nice and save much effort. I need to look some more.

PR #179, opened 2019-11-25
Map aesm service running on host to docker container

manju956:
aesm service manages enclaves running on platform, its not useful to
run aesm service on every instance of docker container.
Hence service running on host is mapping in the form of socket to
docker container (required only in SGX HW mode).

### DIFF
--- a/PREREQUISITES.md
+++ b/PREREQUISITES.md
@@ -196,6 +196,19 @@ wget https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/libsgx-enclave-co
 sudo dpkg -i libsgx-enclave-common_2.3.101.46683-1_amd64.deb
 ```
 
+### Run aesm service on host machine
+If you are behind a corporate proxy,
+uncomment and update the proxy type and aesm proxy lines in /etc/aesmd.conf:  
+```
+proxy type = manual
+aesm proxy = http://your-proxy:your-port
+```
+
+Start aesm service on host machine  
+```
+sudo source /opt/intel/libsgx-enclave-common/aesm/aesm_service
+```
+
 ### Remove Old `/dev/sgx` Intel SGX DCAP Driver
 If device file `/dev/sgx` is present, remove the old DCAP driver:
 ```bash
--- a/docker-compose-sgx.yaml
+++ b/docker-compose-sgx.yaml
@@ -33,6 +33,8 @@ services:
       # Below volume mappings are required for DISPLAY settings by heart disease GUI client
       - /tmp/.X11-unix:/tmp/.X11-unix
       - ~/.Xauthority:/root/.Xauthority
+      # Map aesm service on host to docker, required only in HW mode.
+      - /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
     devices:
       - "/dev/isgx:/dev/isgx"
     working_dir: "/project/TrustedComputeFramework"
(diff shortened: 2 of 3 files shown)


docker-compose-sgx.yaml
```diff
@@ -16,8 +16,8 @@ version: '3.5'
 
 services:
   tcf:
-    container_name: tcf
-    image: tcf-dev
+    container_name: tcf-manju
```
pankajgoyal2: please remove 'manju' from container_name.

docker-compose-sgx.yaml
```diff
@@ -33,6 +33,8 @@ services:
       # Below volume mappings are required for DISPLAY settings by heart disease GUI client
       - /tmp/.X11-unix:/tmp/.X11-unix
       - ~/.Xauthority:/root/.Xauthority
+      # Map aesm service on host to docker, required only in HW mode.
```
pankajgoyal2: can you remove http_proxy, https_proxy, no_proxy arguments. In Dockerfile.tcf-dev, these variables are not being read from here.

docker-compose-sgx.yaml
manju956: ok

docker-compose-sgx.yaml
manju956: yes, proxy variables are actually read from /etc/environment

manju956:
> to install sgx driver on host machine and start the aesm service befor

Yes, we have to install SGX driver and libsgx-enclave-common packages on the host machine and start the aesm service. Since, aesm service manages enclaves that are running on the system its good practice to map host service to docker instead of running aesm service on every instance of docker that initializes enclave.

tools/rebuild.sh
```diff
@@ -75,25 +75,7 @@ fi
...
-    if [ $? -ne 0 -a -n "$http_proxy" ] ; then
-       echo "Setting AESMD proxy"
-       echo "aesm proxy = $http_proxy" >> /etc/aesmd.conf
-    fi
-
-    # Starting aesm service
-    echo "Starting aesm service"
-    /opt/intel/libsgx-enclave-common/aesm/aesm_service &
-else
+if [ -z ${SGX_MODE} ]; then
```
arsulegai: Is it ok to continue if SGX_MODE is set to anything else other than `HW` or `SIM`?

PREREQUISITES.md
```diff
@@ -196,6 +196,19 @@ wget https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/libsgx-enclave-co
...
 
+### Run aesm service on host machine
+If you are behind a corporate proxy,
+uncomment and update the proxy type and aesm proxy lines in /etc/aesmd.conf:  
+```
+proxy type = manual
+aesm proxy = http://your-proxy:your-port
+```
+
+Start aesm service on host machine
```
arsulegai: Can we put these lines in the beginning of this sub-section? The paragraph starting with the if is appearing as if starting the aesmd is an optional.

manju956: I just put the two sections i.e setting proxy and running aesm to follow the sequence.

tools/rebuild.sh
manju956: no modes other than SIM or HW should go through. I will add the check.

PREREQUISITES.md
arsulegai: Line starting with `###` would appear a sub-section. Under that I see 2 paragraphs, starting the `aesm` should be told first and then a note on machines behind the proxy network.

PR #184, opened 2019-12-02
Enable pycodestyle Static Checking on Builds

danintel:


danintel:
FYI, the script `run_lint` is from the sawtooth-core repository.

danintel:
This is a stripped-down version of an earlier PR, #172

PR #186, opened 2019-12-04
Fix pycodestyle Errors for `tests/` and `examples/apps/aml`

danintel:
- Fix pycodestyle errors for `tests/`
- Fix tab errors for
  `examples/apps/aml/standalone-cipher/encryptionAlg.py`

### DIFF
--- a/tests/Demo.py
+++ b/tests/Demo.py
@@ -29,28 +29,27 @@ import utility.utility as enclave_helper
 import utility.file_utils as futils
 from error_code.error_status import SignatureStatus, WorkOrderStatus
 
+LOGGER = logging.getLogger(__name__)
 
 TCFHOME = os.environ.get("TCF_HOME", "../../")
-logger = logging.getLogger(__name__)
 
 
 # -----------------------------------------------------------------
 # -----------------------------------------------------------------
-def LocalMain(config):
-
+def local_main(config):
     if not input_json_str and not input_json_dir:
-        logger.error("JSON input file is not provided")
+        LOGGER.error("JSON input file is not provided")
         exit(1)
 
     if not output_json_file_name:
-        logger.error("JSON output file is not provided")
+        LOGGER.error("JSON output file is not provided")
         exit(1)
 
     if not server_uri:
-        logger.error("Server URI is not provided")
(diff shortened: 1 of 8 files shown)


tests/Demo.py
```diff
@@ -263,17 +277,19 @@ def Main(args=None):
 
     try:
         config = pconfig.parse_configuration_files(conffiles, confpaths)
-        json.dumps(config, indent=4)
+        config_json_str = json.dumps(config, indent=4)
```
manojgop: Looks like config_json_str is not used. Will pylint report unused variable issue ?

tests/Demo.py
```diff
@@ -263,17 +277,19 @@ def Main(args=None):
...
-    sys.stdout = plogger.stream_to_logger(logging.getLogger("STDOUT"), logging.DEBUG)
-    sys.stderr = plogger.stream_to_logger(logging.getLogger("STDERR"), logging.WARN)
+    sys.stdout = plogger.stream_to_logger(logging.getLogger("STDOUT"),
+                                          logging.DEBUG)
+    sys.stderr = plogger.stream_to_logger(logging.getLogger("STDERR"),
+                                          logging.WARN)
 
     ParseCommandLine(config, remainder)
-    LocalMain(config)
+    local_main(config)
```
manojgop: Do we need to change ParseCommandLine -> parse_command_line

tests/Demo.py
danintel: You're right. Looks like it was there for debugging. I commented it out.

tests/Demo.py
danintel: Done. I fixed pycodestyle issues, but not all pylint issues.
That will need to be done later.

tests/Demo.py
manojgop: I think the line " json.dumps(config, indent=4)" was added to check if config is a well-formatted json string else it will throw an exception. So this line was not added just for debugging. So this line shouldn't be commented out.

danintel: FIxed.  The `json.dumps` line is now unchanged.

danintel:
@manojgop Please re-review.

PR #188, opened 2019-12-04
Fix pycodestyle Errors for `examples/enclave_manager/`

danintel:
- Fix pycodestyle errors
- Enable run_lint for `examples/enclave_manager`

### DIFF
--- a/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py
+++ b/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py
@@ -30,7 +30,8 @@ import crypto.crypto as crypto
 from database import connector
 from error_code.error_status import ReceiptCreateStatus, WorkOrderStatus
 from utility.tcf_types import WorkerStatus, WorkerType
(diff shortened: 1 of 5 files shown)


examples/enclave_manager/tcf_enclave_manager/enclave_manager.py
```diff
@@ -30,7 +30,8 @@
 from database import connector
 from error_code.error_status import ReceiptCreateStatus, WorkOrderStatus
 from utility.tcf_types import WorkerStatus, WorkerType
-from work_order_receipt.work_order_receipt_request import WorkOrderReceiptRequest
+from work_order_receipt.work_order_receipt_request \
+    import WorkOrderReceiptRequest
 
 logger = logging.getLogger(__name__)
```
manojgop: I saw in an earlier PR logger renamed to LOGGER

danintel: Yes, I renamed it to `LOGGER` because pylint3 marked that as an error.
But running pylint3 requires too many changes to make right now.

PR #190, opened 2019-12-06
Fix fread() of File Opened Write-only

danintel:
Found running `cppcheck --std=c++19 $(find . -name *cpp)`

Problem is in `tc/sgx/common/tests/trusted/app/TestApp.cpp`
where an empty file is created then read.


danintel:
Note: I added cppcheck to run_lint in a separate PR, #191, but it's not dependent on that PR.

PR #191, opened 2019-12-07
Add cppcheck to `run_lint`

danintel:
- Add cppcheck for static C++ source code checking to run_lint
- Add cppcheck package to Docker
- Fix `docker/Dockerfile.tcf-dev` build error by using `ln -sf` :
`failed to create symbolic link '/usr/bin/python': File exists`

Signed-off-by: danintel <daniel.anderson@intel.com>

PR #193, opened 2019-12-12
Fix syntax error in test_cryptoWrapper.py

danintel:
Fix syntax error in test_cryptoWrapper.py

- Remove extra double quote


danintel:
Followup to #191

PR #201, opened 2020-01-09
Build static libraries for base64 and parson packages

manju956:
Previously both base64 and parson packages used to build as part of tc/sgx/common library.
Due to this, any module which needs to used them has to build bulk sized common library.
Seperating these packages into libraries make them light weight.

### DIFF
--- a/tc/sgx/common/packages/base64/CMakeLists.txt
+++ b/tc/sgx/common/packages/base64/CMakeLists.txt
@@ -1,0 +1,62 @@
+# Copyright 2019 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+CMAKE_MINIMUM_REQUIRED(VERSION 3.2 FATAL_ERROR)
+find_package(PkgConfig REQUIRED)
+
+option(UNTRUSTED_ONLY "Build only untrusted components" OFF)
+
+################################################################################
+# Common components for both trusted and untrusted base64 libraries
+################################################################################
+
+FILE(GLOB PROJECT_HEADERS *.h)
+FILE(GLOB PROJECT_SOURCES *.cpp)
+
+SET(COMMON_CXX_FLAGS ${DEBUG_FLAGS} "-m64" "-fvisibility=hidden" "-fpie" "-fPIC" "-fstack-protector" "-std=c++11" "-Wall")
+
+################################################################################
+# Untrusted Base64 Library
+################################################################################
+
+SET(UNTRUSTED_LIB_NAME uavalon-base64)
+PROJECT(${UNTRUSTED_LIB_NAME} CXX)
+
+ADD_LIBRARY(${UNTRUSTED_LIB_NAME} STATIC ${PROJECT_HEADERS} ${PROJECT_SOURCES})
+
+TARGET_COMPILE_OPTIONS(${UNTRUSTED_LIB_NAME} PRIVATE ${COMMON_CXX_FLAGS})
+
+TARGET_COMPILE_DEFINITIONS(${UNTRUSTED_LIB_NAME} PRIVATE "-D_UNTRUSTED_=1")
+
+################################################################################
+# Trusted Base64 Library
+################################################################################
+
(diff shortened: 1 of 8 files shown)


tc/sgx/common/packages/base64/base64.h
```diff
@@ -9,12 +9,14 @@
 
 #pragma once
 
-#include "types.h"
+#include <string>
+#include <vector>
+//#include "types.h"
 
 #define BASE64_SIZE(x) (static_cast<size_t>(((((x) - 1) / 3) * 4 + 4) + 1))
```
danintel: I do not see this used anywhere.
Also it will not work with types.h commented out.
I would delete it.
If you want to keep it, add a comment about what it does (length of a base 64 encoded string for a plaintext string? Including or excluding trailing NULL character?)

tc/sgx/common/packages/base64/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+
+################################################################################
+# Common components for both trusted and untrusted base64 libraries
+################################################################################
+
+FILE(GLOB PROJECT_HEADERS *.h)
+FILE(GLOB PROJECT_SOURCES *.cpp)
+
+SET(COMMON_CXX_FLAGS ${DEBUG_FLAGS} "-m64" "-fvisibility=hidden" "-fpie" "-fPIC" "-fstack-protector" "-std=c++11" "-Wall")
+INCLUDE_DIRECTORIES()
```
manojgop: Do we need empty INCLUDE_DIRECTORIES() ?

manju956: No. I will delete it

tc/sgx/common/packages/base64/base64.h
manju956: I will delete BASE64_SIZE

tc/sgx/common/packages/parson/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+# Copyright 2019 Intel Corporation
```
arsulegai: 2020?

tc/sgx/common/packages/base64/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+# Copyright 2019 Intel Corporation
```
arsulegai: 2020?

tc/sgx/common/packages/base64/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+FILE(GLOB PROJECT_HEADERS *.h)
+FILE(GLOB PROJECT_SOURCES *.cpp)
+
+SET(COMMON_CXX_FLAGS ${DEBUG_FLAGS} "-m64" "-fvisibility=hidden" "-fpie" "-fPIC" "-fstack-protector" "-std=c++11" "-Wall")
+
+################################################################################
+# Untrusted Base64 Library
+################################################################################
+
+SET(UNTRUSTED_LIB_NAME utcf-base64)
```
arsulegai: Curious question, are we still following xtcf-xxxx convention or it can be moved to xavalon-xxx?

tc/sgx/common/packages/base64/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+################################################################################
+
+SET(UNTRUSTED_LIB_NAME uavalon-base64)
+PROJECT(${UNTRUSTED_LIB_NAME} CXX)
+
+ADD_LIBRARY(${UNTRUSTED_LIB_NAME} STATIC ${PROJECT_HEADERS} ${PROJECT_SOURCES})
+
+TARGET_COMPILE_OPTIONS(${UNTRUSTED_LIB_NAME} PRIVATE ${COMMON_CXX_FLAGS})
+
+TARGET_COMPILE_DEFINITIONS(${UNTRUSTED_LIB_NAME} PRIVATE "-D_UNTRUSTED_=1")
```
arsulegai: Preferably define a constant variable outside to pass any definition here and other places in these cmake files.

tc/sgx/common/packages/base64/CMakeLists.txt
arsulegai: I see it is changed before posting the comment.

tc/sgx/common/packages/parson/CMakeLists.txt
manju956: I will take care of updating copyright for these files in my new PR.

tc/sgx/common/packages/base64/CMakeLists.txt
manju956: I will take care of updating copyright for these files in my new PR.

tc/sgx/common/packages/base64/CMakeLists.txt
manju956: Already taken care.

PR #203, opened 2020-01-10
Separate SGX error handling into a new file

manju956:
This change is done to avoid SGX dependency on any module which just needs
to use general error handling functionalities provided by error.h

### DIFF
--- a/tc/sgx/common/error.h
+++ b/tc/sgx/common/error.h
@@ -19,32 +19,9 @@
 #include <string>
 
 #include "tcf_error.h"
-#include "sgx_error.h"
 
 namespace tcf {
     namespace error {
-
-        // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-        class SgxError : public std::exception {
-        public:
-            explicit SgxError() : errorCode(SGX_SUCCESS) {}
-            explicit SgxError(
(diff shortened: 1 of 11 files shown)


tc/sgx/trusted_worker_manager/enclave/base_enclave.cpp
```diff
@@ -28,6 +28,7 @@
 #include "auto_handle_sgx.h"
 
 #include "error.h"
+#include "avalon_sgx_error.h"
 #include "tcf_error.h"
```
manojgop: We can combine error.h and tcf_error.h into avalon_error.h. May be in another PR ?

manju956: Yes, i will combine them in another PR.

tc/sgx/common/packages/base64/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Intel Corporation
+# Copyright 2020 Intel Corporation
```
manojgop: Not a related change for SGX error handling. But fine as it's a minor change updating copyright year.

tc/sgx/common/error.h
```diff
@@ -23,28 +23,6 @@
```
manojgop: sgx_error.h inclusion has to be removed

manju956: Done

PR #210, opened 2020-01-22
Fix Error `error: invalid command 'bdist_wheel'`

danintel:
The following is needed to fix this error from running `python setup.py bdist_wheel:

`error: invalid command 'bdist_wheel'`

### DIFF
--- a/PREREQUISITES.md
+++ b/PREREQUISITES.md
@@ -86,9 +86,9 @@ sudo apt-get install -y cmake swig pkg-config python3-dev python3-venv python \
      python3-requests python3-colorlog python3-twisted
 ```
 
-Also, install the following pip packages
+Also, install the following pip3 packages:
 ```bash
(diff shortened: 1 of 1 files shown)


examples/common/python/Makefile
```diff
@@ -24,6 +24,7 @@ all : $(WHEEL_FILE)
 
 $(WHEEL_FILE) : build_ext
 	@echo Build Distribution
+	pip3 install wheel
```
Ram-srini: Is this change to fix the stand-alone build?

danintel: @Ram-srini : Yes

danintel: I will remove the `pip3 install wheel` and modify instead the PREREQUISITES.md to install pip3 as non-root. This should fix this problem and other problems.

If the pip3 packages also need to be installed as root, let me know.

PR #211, opened 2020-01-22
Modify generic client to use avalon client sdk functions

manojgop:
- Use Avalon client sdk functions in generic client
- rename utility to crypto_utils

### DIFF
--- a/examples/apps/generic_client/generic_client.py
+++ b/examples/apps/generic_client/generic_client.py
@@ -23,12 +23,18 @@ import secrets
 
 import config.config as pconfig
 import utility.logger as plogger
-import crypto_utils.crypto_utility as utility
+import crypto_utils.crypto_utility as crypto_utility
 from avalon_client_sdk.utility.tcf_types import WorkerType
 import avalon_client_sdk.worker.worker_details as worker_details
 from avalon_client_sdk.work_order.work_order_params import WorkOrderParams
-from avalon_client_sdk.direct.avalon_direct_client \
-    import AvalonDirectClient
+from avalon_client_sdk.ethereum.ethereum_worker_registry_list_client import \
+    EthereumWorkerRegistryListClientImpl
+from avalon_client_sdk.direct.jrpc.worker_registry_jrpc_client import \
+    WorkerRegistryJRPCClientImpl
+from avalon_client_sdk.direct.jrpc.work_order_jrpc_client import \
+    WorkOrderJRPCClientImpl
+from avalon_client_sdk.direct.jrpc.work_order_receipt_jrpc_client \
+     import WorkOrderReceiptJRPCClientImpl
 from error_code.error_status import WorkOrderStatus, ReceiptCreateStatus
 import crypto_utils.signature as signature
 from error_code.error_status import SignatureStatus
@@ -42,19 +48,7 @@ logger = logging.getLogger(__name__)
 TCFHOME = os.environ.get("TCF_HOME", "../../")
 
 
-def ParseCommandLine(args):
-
-    global worker_obj
-    global worker_id
-    global workload_id
-    global in_data
-    global config
-    global mode
-    global uri
-    global address
-    global show_receipt
-    global show_decrypted_output
-    global requester_signature
+def _parse_command_line(args):
 
     parser = argparse.ArgumentParser()
     mutually_excl_group = parser.add_mutually_exclusive_group()
@@ -105,8 +99,13 @@ def ParseCommandLine(args):
         action="store_true")
     options = parser.parse_args(args)
 
-    if options.config:
-        conf_files = [options.config]
+    return options
+
+
+def _parse_config_file(config_file):
+    # Parse config file and return a config dictionary.
+    if config_file:
+        conf_files = [config_file]
     else:
         conf_files = [TCFHOME +
                       "/client_sdk/avalon_client_sdk/tcf_connector.toml"]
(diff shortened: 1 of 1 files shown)


examples/apps/generic_client/generic_client.py
```diff
@@ -165,9 +171,6 @@ def Main(args=None):
     logger.info("***************** TRUSTED COMPUTE FRAMEWORK (TCF)" +
```
pankajgoyal2: TRUSTED COMPUTE FRAMEWORK (TCF)  => Avalon

examples/apps/generic_client/generic_client.py
```diff
@@ -165,9 +171,6 @@ def Main(args=None):
...
                 " *****************")
 
-    global direct_jrpc
-    direct_jrpc = AvalonDirectClient(config_file=None, config=config)
-
     global address
```
pankajgoyal2: why do we have so many global variables defined. This way of writing code is not good at all. It looks that demo.py has been referenced while developing this file. Please take action on this cleanup. This is my comment. Seperate PR can be raised for that cleanup. I'm not making this as blocking for this PR.

manojgop: Yes. This issue was identified earlier also. This can be fixed as a separate PR. I'll raise an issue in github

examples/apps/generic_client/generic_client.py
manojgop: ok

examples/apps/generic_client/generic_client.py
```diff
@@ -3,7 +3,7 @@
 # Copyright 2019 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
+# you may not use this /ile except in compliance with the License.
```
danintel: s:/ile:file:

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+    logger.info("\n Retrieve Http JSON RPC listener uri from registry \n")
+    # Get block chain type
+    blockchain_type = config['blockchain']['type']
+    if blockchain_type == "Ethereum":
+            worker_registry_list = EthereumWorkerRegistryListClientImpl(
+                config)
+    else:
+        worker_registry_list = None
+        logger.error("\n Worker registry list is currently supported only for \
+            ethereum block chain \n")
```
danintel: This results in embedded spaces in the error message. Instead use:

`logger.error("\n Worker registry list is currently supported only for "`
`            "ethereum block chain \n")`

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+        return None
+
+    # Lookup returns tuple, first element is number of registries and
+    # second is element is lookup tag and
+    # third is list of organization ids.
+    registry_count, lookup_tag, registry_list = \
+        worker_registry_list.registry_lookup()
+    logger.info("\n Registry lookup response: registry count: {} " +
```
danintel: Same as above.

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+                "lookup tag: {} registry list: {}\n".format(
+                    registry_count, lookup_tag, registry_list))
+    if (registry_count == 0):
+        logger.error("No registries found")
+        return None
+    # Retrieve the first registry details.
+    registry_retrieve_result = worker_registry_list.registry_retrieve(
+        registry_list[0])
+    logger.info("\n Registry retrieve response: {}\n".format(
+        registry_retrieve_result
```
danintel: Same as above.

examples/apps/generic_client/generic_client.py
manojgop: ok

examples/apps/generic_client/generic_client.py
manojgop: I addressed this in this PR itself

examples/apps/generic_client/generic_client.py
manojgop: ok

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+    ))
+
+    return registry_retrieve_result[0]
+
+
+def _lookup_first_worker(worker_registry, jrpc_req_id):
```
Ram-srini: Is the method name _lookup_first_worker make sense if we go for multiple workers in future. Change it to lookup_worker and mention comment saying currently one is supported and getting first worker.

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+            logger.error("ERROR: No workers found")
+            worker_id = None
+    else:
+        logger.error("ERROR: Failed to lookup worker")
+        worker_id = None
+
+    return worker_id
+
+
+def _create_work_order(worker_id, workload_id, in_data, worker_encrypt_key,
```
Ram-srini: It is creating work order request not submitting request. change it to _create_work_order =>_create_work_order_request

examples/apps/generic_client/generic_client.py
```diff
@@ -289,66 +436,37 @@ def Main(args=None):
...
-            wo_create_receipt["requesterSignature"],
-            wo_create_receipt["signatureRules"],
-            wo_create_receipt["receiptVerificationKey"],
-            req_id
-        )
-        logger.info("Work order create receipt response : {} \n \n ".format(
-            wo_receipt_resp
-        ))
+        jrpc_req_id += 1
+        _create_work_order_receipt(wo_receipt, wo_params,
```
Ram-srini: _create_work_order_receipt() is doing creation of receipt and submitting request but _create_work_order() is creating a only request. Either move work order submission part also to _create_work_order() or change other APIs to do the same.

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+        return False
+
+    return True
+
+
+def _verify_wo_res_signature(work_order_res,
+                             worker_verification_key):
+    sig_obj = signature.ClientSignature()
+    status = sig_obj.verify_signature(work_order_res, worker_verification_key)
+    try:
```
Ram-srini: Why there try-catch for logging?

examples/apps/generic_client/generic_client.py
```diff
@@ -116,14 +115,203 @@ def ParseCommandLine(args):
...
+        json.dumps(receipt_update_retrieve, indent=4)
+    ))
+
+    return receipt_update_retrieve
+
+
+def _verify_receipt_signature(receipt_update_retrieve):
+    # Verify receipt signature
+    sig_obj = signature.ClientSignature()
+    status = sig_obj.verify_update_receipt_signature(
```
Ram-srini: Do need try-catch here?

examples/apps/generic_client/generic_client.py
manojgop: This function is to choose first worker as default worker if worker_id is not provided as an argument to generic client. I don't except this default behavior to change any time soon.

examples/apps/generic_client/generic_client.py
manojgop: or should we rename it to _create_work_order_params ? This function is returning WorkOrderParams object

examples/apps/generic_client/generic_client.py
manojgop: The  implementation of verify_update_receipt_signature( has try/catch. So here only function return value is checked.

examples/apps/generic_client/generic_client.py
manojgop: No need. We can remove it

examples/apps/generic_client/generic_client.py
manojgop: will change _create_work_order -> _create_work_order_param. _create_work_order_receipt internally calls " work_order_receipt_create" function in https://github.com/hyperledger/avalon/blob/master/client_sdk/avalon_client_sdk/direct/jrpc/work_order_receipt_jrpc_client.py#L36. So to maintain consistency used this function name. We can add a comment in the function that this will send Work Order Receipt json rpc request

PR #212, opened 2020-01-22
Disable echo client run in CI tests

Ram-srini:
Infura Ethereum provider is moved to new http interface, this needs some changes to ethereum wrapper.
Until we fix that temporarily echo client test is disabled from CI run.


pankajgoyal2:
Completely disabling echo client tests from CI is not a good idea. Probably what can be done is to have an option in echo client to bypass the blockchain and run the tests. I am approving this now but no major PR should be merged until we get and update the new http interface URL.

PR #214, opened 2020-01-22
tcs_startup.sh: Add LMDB Option and Change Name to Hyperledger Avalon

danintel:
Add option to specify LMDB server.  For example:
        tcs_startup.sh -l http://avalon-lmdb:9090
Default to http://localhost:9090

Change TCS or TCF to Hyperledger Avalon

### DIFF
--- a/scripts/tcs_startup.sh
+++ b/scripts/tcs_startup.sh
@@ -14,30 +14,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-enclave_manager="${TCF_HOME}/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py"
-listener="${TCF_HOME}/common/python/connectors/direct/tcs_listener/tcs_listener.py"
-version="$(cat ${TCF_HOME}/VERSION)"
+ENCLAVE_MANAGER="${TCF_HOME}/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py"
+LISTENER="${TCF_HOME}/common/python/connectors/direct/tcs_listener/tcs_listener.py"
+VERSION="$(cat ${TCF_HOME}/VERSION)"
+# Default values
+LMDB_URL="http://localhost:9090" # -l default
 
 # Trap handler
-trap 'stop_tcs_components' HUP INT QUIT ABRT ALRM TERM
+trap 'stop_avalon_components' HUP INT QUIT ABRT ALRM TERM
 
-start_tcs_components()
+start_avalon_components()
 {
     # Read Listener port from config file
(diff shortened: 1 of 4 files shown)


scripts/tcs_startup.sh
```diff
@@ -14,66 +14,73 @@
...
-trap 'stop_tcs_components' HUP INT QUIT ABRT ALRM TERM
+trap 'stop_avalon_components' HUP INT QUIT ABRT ALRM TERM
 
-start_tcs_components()
+start_avalon_components()
 {
     # Read Listener port from config file
-    listener_port=`grep listener_port ${TCF_HOME}/config/tcs_config.toml \
+    LISTENER_PORT=`grep listener_port ${TCF_HOME}/config/tcs_config.toml \
         | awk {'print $3'}`
```
Ram-srini: Where are we starting the kv_storage process in standalone run? We shouldn't include that in tcs_startup.sh because it will impact the docker based build. We should have a separate shell script to start the kv_storage process.

scripts/tcs_startup.sh
```diff
@@ -14,66 +14,73 @@
...
-    if [ -z "$listener_port" ]; then
-        listener_port=1947
+    if [ -z "$LISTENER_PORT" ]; then
+        LISTENER_PORT=1947
     fi
 
-    echo "Starting Avalon Enclave Manager $version ..."
-    python3 $enclave_manager --lmdb_url http://avalon-lmdb:9090 &
```
Ram-srini: Also we need to document to start the kv_storage service has to be started first.

scripts/tcs_startup.sh
danintel: I will file a separate PR to add that unrelated feature.

I am just trying to fix this broken script so it will work in standalone mode.

scripts/tcs_startup.sh
danintel: Ok

scripts/tcs_startup.sh
danintel: @Ram-srini I filed a separate PR #217 to add `kv_storage` startup, which apparently was not added or documented before.  Please review any kv_storage related issues in that PR.

Ram-srini: @danintel So Does it means #217 is one we should consider not the PR#214 right?

danintel:
Please consider PR #214. It helps reduce the breakage and regression in Avalon.

Dan

scripts/tcs_startup.sh
danintel: @Ram-srini Please consider both pull requests. PR #217 will be partly reduced once this initial base PR is approved. I cannot merge this PR until you approve this PR and unblock it.

PR #217, opened 2020-01-23
tcs_startup.sh: Add kv_storage startup

danintel:
tcs_startup.sh: Add kv_storage startup

tcs_startup.sh
- Start kv_storage before everything else:
   `kv_storage --bind http://avalon-lmdb:9090`
   (or the value provided by the `-l` option)
- Add `-s` option to optionally not start kv_storage

Dockerfiles
- Add `-s` option when caling tcs_startup.sh from Docker

### DIFF
--- a/ci/tcf-tests.yaml
+++ b/ci/tcf-tests.yaml
@@ -25,7 +25,7 @@ services:
         - no_proxy
     working_dir: "/project/TrustedComputeFramework/"
     entrypoint: "bash -c \"\
-        source scripts/tcs_startup.sh -y \
+        source scripts/tcs_startup.sh -y -s \
         && sleep 5 \
         && source tools/run_tests.sh \""
     depends_on:
--- a/scripts/tcs_startup.sh
+++ b/scripts/tcs_startup.sh
@@ -14,9 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+KV_STORAGE="kv_storage"
 enclave_manager="${TCF_HOME}/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py"
 listener="${TCF_HOME}/common/python/connectors/direct/tcs_listener/tcs_listener.py"
 version="$(cat ${TCF_HOME}/VERSION)"
+# Default values
+START_KV_STORAGE=1 # default if -s not passed
 
 # Trap handler
 trap 'stop_tcs_components' HUP INT QUIT ABRT ALRM TERM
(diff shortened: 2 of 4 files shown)


scripts/tcs_startup.sh
```diff
@@ -14,30 +14,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+KV_STORAGE="$TCF_HOME/tools/build/_dev/bin/kv_storage"
 enclave_manager="${TCF_HOME}/examples/enclave_manager/tcf_enclave_manager/enclave_manager.py"
 listener="${TCF_HOME}/common/python/connectors/direct/tcs_listener/tcs_listener.py"
 version="$(cat ${TCF_HOME}/VERSION)"
+# Default values
+DEFAULT_LISTENER_PORT=1947 # set in config/tcs_config.toml
+START_KV_STORAGE=1 # default if -s not passed
```
Ram-srini: Where is "-s" parameter is checked in the script? By default START_KV_STORAGE is set 1 and it is starting kv_storage service in tcf docker container as well.

manojgop: Looks fine to me. "-s" option is passed with tcs_startup.sh in docker compose file. So KV storage won't start.

manojgop: Another option is to remove the "-s" option from this script and start kv storage separately in stand alone build

ci/tcf-tests.yaml
```diff
@@ -25,7 +25,7 @@ services:
         - no_proxy
     working_dir: "/project/TrustedComputeFramework/"
     entrypoint: "bash -c \"\
-        source scripts/tcs_startup.sh -y \
+        source scripts/tcs_startup.sh -y -s -l http://avalon-lmdb:9090 \
```
pankajgoyal2: Why don't you remove running kv_storage service from tcs_startup.sh at all. Now, this is a seperate service, be it standalone or docker mode, we should start this service seperately.

scripts/tcs_startup.sh
danintel: @Ram-srini: The `-s` option is parsed by `getopts` around line 75.
The `-s` option is passed from docker com a Dockerfile as mentioned by @manojgop.

@manojgop Yes, `kv_storage` can be started separately and that will work. But the purpose of `tcs_startup.sh` is to simplify things and start all the components.

ci/tcf-tests.yaml
danintel: @Ram-srini: consider both PR #214 and #217. PR #217 builds on the changes in PR #214 These PRs are urgently needed to fix the regressions in Avalon so I can deliver a working version of Avalon that is needed by a partner who wants to use Avalon.

@pankajgoyal2 `tcs_startup.sh` starts all the services needed by Avalon in one simple startup script, including the enclave manager and listener. Avalon is very complex and there is no need to make it more difficult to use.

danintel: @Ram-srini: to reduce confusion I removed the duplicate changes from PR #214.  Please consider both PR #214 and this PR separately.  I will merge after approval.

scripts/tcs_startup.sh
Ram-srini: Got it. Thanks.

danintel:
I'm closing this PR and opening a new PR #223 .
PR #223 is different in that `kv_storage` is started only if `tcs_startup.sh -s` is used, so it should be less controversial.

PR #218, opened 2020-01-24
Fix Docker Syntax Error

danintel:
Fix Docker Syntax Error

File `docker/Dockerfile.tcf-dev

`08:42:32  Step 34/34 : RUN if [ $MAKECLEAN != 0 ]; then echo "Clean build" && make clean; fi  && make  && echo "BUILD SUCCESS"`
`08:42:33 /bin/sh: 1: [: !=: unexpected operator`

The problem is `$MAKECLEAN` is not surrounded by double quotes.
If $MAKECLEAN is not defined, the if expression becomes a syntax error: `if [ != 0 ] ...`
As a result a `make clean` is never ran.

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -144,7 +144,7 @@ ARG TCF_DEBUG_BUILD
 ARG DISPLAY
 ARG XAUTHORITY
 
-#TODO need to be removed latter
+#TODO need to be removed later
 #install avalon-client python package.
 RUN cd /project/TrustedComputeFramework/client_sdk \
    && python setup.py bdist_wheel \
@@ -192,6 +192,6 @@ WORKDIR /project/TrustedComputeFramework/tools/build
 
 RUN make lint && echo "LINT SUCCESS"
 
-RUN if [ $MAKECLEAN != 0 ]; then echo "Clean build" && make clean; fi \
+RUN if [ "$MAKECLEAN" != 0 ]; then echo "Clean build" && make clean; fi \
  && make \
  && echo "BUILD SUCCESS"


docker/Dockerfile.tcf-dev
```diff
@@ -192,6 +192,6 @@ WORKDIR /project/TrustedComputeFramework/tools/build
 
 RUN make lint && echo "LINT SUCCESS"
 
-RUN if [ $MAKECLEAN != 0 ]; then echo "Clean build" && make clean; fi \
+RUN if [ "$MAKECLEAN" != 0 ]; then echo "Clean build" && make clean; fi \
```
arsulegai: I am curious here! Do we need such a condition in Dockerfile, that too as part of RUN command?

`RUN` command gets executed at docker image creation time. If we need a new build, we anyway will get a clean build everytime. Basically the code will be copied to the new intermediate image before executing make again. I do not see need for this condition in the Dockerfile. If we do not want previous builds getting copied during the `COPY . /folder/in/image`, then those folders/files can be added to `.dockerignore` file.

manojgop: @arsulegai That looks like a valid point. @manju956 @pankajgoyal2 Could you please check if we need to do make clean in dockerfile.

manju956: MAKECLEAN flag is added to skip building cpp modules with CMAKE files and to build only python modules (MAKECLEAN=0). This would save time if there are no changes being made in cpp modules.
By the way since we are not volume mapping previous build files doesn't reflect on host and there is no issue of build files getting copied to docker context in the subsequent docker builds.

arsulegai: I see `make clean` command executed when the flag `MAKECLEAN` is set to `0`. Explanation here is that the flag is used to skip building a part of the module. Multiple ambiguities here!

Here are my comments
1. Change the flag's name to SKIP_CPP or something that matches to the explanation given here.
2. At least perform `make clean` operation when a variable something like `MAKECLEAN` is set (non 0 value).
3. Ideally this should be a `Makefile` (a set of commands to build the code based on condition) feature and not a condition to be checked outside the `Makefile`.

Request you to give more clarity to the code.

manju956: Currently, make clean target is executed if MAKECLEAN set to non 0 or if no value if being passed.
I agree with you on renaming the flag to SKIP_CPP_MODULES which would give more clarity.
Also, IMO the place holder for checking the flag should be in docker file from where top level make is called. because the make targets needs to be passed outside of Makefile

arsulegai: Thanks for the `!=` part, I missed that one. But yeah, other questions/comments hold true.

PR #223, opened 2020-01-27
tcs_startup.sh: Add Option to Start kv_storage

danintel:
Add `-s` option to optionally start kv_storage
before other Avalon components.


pankajgoyal2:
Soon we are going to have listener as seperate service. We need to handle that scenario like this?

PR #228, opened 2020-01-29
PREREQUISITES.md: Add Note About conflicting `threads.h` types

danintel:
In `threads.h`, both `pthread_key_t` and `pthread_once_t` should be
`unsigned int`, not `int`.  If not, these build errors occur:

`error: conflicting types for ‘pthread_key_t’`
`error: conflicting types for ‘pthread_once_t’`

The solution is to use the current version of the Intel SGX SDK.


danintel:
Merged with #227, so closing this PR.

PR #236, opened 2020-02-03
Fix Lint Checks

danintel:
- Add Python checks for `common/`
- Fix cppheck for `#ifdef configurations`
- Fix Python lint errors

### DIFF
--- a/common/python/crypto_utils/signature.py
+++ b/common/python/crypto_utils/signature.py
@@ -111,7 +111,8 @@ class ClientSignature(object):
             e_key = item['encryptedDataEncryptionKey'].encode('UTF-8')
 
             if (not e_key) or (e_key == "null".encode('UTF-8')):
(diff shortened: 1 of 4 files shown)


common/python/crypto_utils/signature.py
```diff
@@ -111,7 +111,8 @@ def __encrypt_workorder_indata(
             e_key = item['encryptedDataEncryptionKey'].encode('UTF-8')
 
             if (not e_key) or (e_key == "null".encode('UTF-8')):
-                enc_data = crypto_utility.encrypt_data(data, session_key, session_iv)
+                enc_data = crypto_utility.encrypt_data(
```
Ram-srini: Do we need an empty line here?

danintel: @Ram-srini My guess is the line was split into two lines because it was too long.

PR #237, opened 2020-02-03
Fix Tutorial Document

danintel:
Change `bash docker exec -it tcf bash` to
`docker exec -it tcf bash`.


danintel:
Rendered document with correction is here:
https://github.com/danintel/avalon/tree/tutorial/docs/workload-tutorial

PR #238, opened 2020-02-04
tcf_startup.sh: Does Not Terminate When A Daemon Terminates

danintel:
The tcf_startup.sh should terminate if a daemon does not start
or if it terminates. Instead it prints this
forever:

`If you wish to exit the program, press y and enter`


scripts/tcs_startup.sh
```diff
@@ -50,34 +51,47 @@ start_avalon_components()
...
     fi
 }
 
+check_avalon_components()
+{
+    for i in $COMPONENTS ; do
+        pgrep -f "$i"
+        if [ $? != 0 ] ; then
+            echo "Terminating Avalon because component not running:"
+            echo "$LISTENER"
```
manojgop: Why are we only printing $LISTENER here? Is it for debugging purpose ?

danintel: It is reporting an error. The listener terminated for some reason and it is reporting that fact

danintel: My mistake. I fixed it by changing $LISTENER  to $i

PR #240, opened 2020-02-04
Fix Build Warnings

danintel:
`docker/Dockerfile.tcf-dev`
- Remove empty line after continuation line
- s/latter/later/

`tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/enclave.h`
- Remove unused static function `Worker()`

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -151,7 +151,7 @@ RUN cd /project/TrustedComputeFramework/client_sdk \
    && python setup.py bdist_wheel \
    && pip3 install dist/*.whl
 
-#TODO need to be removed latter
+#TODO need to be removed later
 #install avalon-connect sdk python package
 RUN cd /project/TrustedComputeFramework/connector_sdk \
     && python setup.py bdist_wheel \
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.tcf-dev
```diff
@@ -160,8 +160,8 @@ RUN cd /project/TrustedComputeFramework/connector_sdk \
 # Environment setup
 
 RUN echo "WEB3_INFURA_PROJECT_ID=9af17e6320414af8a3168cd9fe9d653d" >> /etc/environment \
-    && echo "WALLET_PRIVATE_KEY=4F611197A6E82715F4D2446FE015D1667E9C40A351411F3A7300F71F285D01B4" >> /etc/environment \
-
+ && echo "WALLET_PRIVATE_KEY=4F611197A6E82715F4D2446FE015D1667E9C40A351411F3A7300F71F285D01B4" \
+    >> /etc/environment \
```
manojgop: Both WEB3_INFURA_PROJECT_ID and WALLET_PRIVATE_KEY could be moved to a config file. May be in a separate PR.

danintel: Both are external to Avalon, so it would be better to not move it to an Avalon-specific file.

But these are separate issues from fixing a syntax error in the Dockerfile.

manojgop: There a config file for block chain connector. So better to move there as a separate PR. @Ram-srini @rranjan3

danintel:
To make this easier to review I split this into 2 PRs:
#246 
#247

PR #242, opened 2020-02-05
FAQ Changes

danintel:
- Add video links
- Add missing feedback from PR #233
  - Intel SGX enclaves
  - Enclave size


danintel:
The rendered FAQ is here (which is easier to read than RST markup):
https://github.com/danintel/avalon/blob/faq/docs/FAQ.rst

danintel:
See #248.

PR #246, opened 2020-02-06
Fix Docker Build Warning

danintel:
`docker/Dockerfile.tcf-dev`
- Remove empty line after continuation line
- Align if/else/fi statements for readability
- s/latter/later/

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -145,7 +145,7 @@ ARG TCF_DEBUG_BUILD
 ARG DISPLAY
 ARG XAUTHORITY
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.tcf-dev
```diff
@@ -145,35 +145,41 @@ ARG TCF_DEBUG_BUILD
 ARG DISPLAY
 ARG XAUTHORITY
 
-#TODO need to be removed latter
+#TODO need to be removed later
```
manojgop: I guess this TODO line can be removed. Not sure why is this still here? @pankajgoyal2 @Ram-srini ?

PR #253, opened 2020-02-07
Adding some check in rebuild.sh

pankajgoyal2:

### DIFF
--- a/tools/rebuild.sh
+++ b/tools/rebuild.sh
@@ -60,6 +60,12 @@ function try() {
     "$@" || die "test failed: $*"
 }
 
+function error_exit()
+{
+    echo "[Error]: " ${*}
+    exit 1
+}
+
 # -----------------------------------------------------------------
 # CHECK ENVIRONMENT
 # -----------------------------------------------------------------
@@ -133,7 +139,7 @@ NUM_CORES=1
 # -----------------------------------------------------------------
 
 yell --------------- "COMMON SGX (WORKLOAD & IOHANDLER)" ---------------
-cd $TCF_HOME/common/sgx_workload/
+cd $TCF_HOME/common/sgx_workload/ || error_exit "Failed to change to the directory"
 
 mkdir -p build
 cd build
@@ -141,7 +147,7 @@ try cmake ..
 try make "-j$NUM_CORES"
 
 yell --------------- EXAMPLE WORKLOADS ---------------
-cd $TCF_HOME/examples/apps
+cd $TCF_HOME/examples/apps || error_exit "Failed to change to the directory"
 
 mkdir -p build
 cd build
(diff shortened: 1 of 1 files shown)


tools/rebuild.sh
```diff
@@ -60,6 +60,14 @@ function try() {
...
 }
 
+check_error()
+{
+    if [ "$?" = "1" ]; then
+        echo "Directory doesn't exist!" 1>&2
+        exit 1
+    fi
+}
+
```
danintel: Good idea--it would make diagnosing problems easier.

This could be simpler if we put both `cd` and the error check into 1 function. Then instead of:
```
cd $TCF_HOME/common/sgx_workload/
check_error
```
You only need:
`cd_check $TCF_HOME/common/sgx_workload/`

The check should check the `$?` status to not be 0--in case cd has other exit codes besides 0 or 1.
Also generalize the message.  The user may not have execute (`x`) permission to `cd` to the directory.

The `cd_check` function could be something like:
```
cd_check()
{
    if [ ! -d "$1" ] ; then
        echo "Directory does not exist: $1" 1>&2
        exit 1
    fi

   cd $1
    if [ "$?" != "0" ]; then
        echo "Cannot change to directory $1" 1>&2
        exit 1
    fi
}
```

danintel:
Looks like the build fails.  I think it's a false failure.  I notice if two builds are occurring at the same time from PRs, one fails and restarting the build (with a `push -f` of your branch for example) fixes the problem.

pankajgoyal2:
> Looks like the build fails. I think it's a false failure. I notice if two builds are occurring at the same time from PRs, one fails and restarting the build (with a `push -f` of your branch for example) fixes the problem.

Yes, you are right. It fails because the compose file defines container_name for avalon-lmdb service in https://github.com/hyperledger/avalon/blob/master/ci/tcf-tests.yaml#L35. Because of this, if parallel jobs run for same PR, all will try to use the same name container and will have the issue. If we don't specify the container_name, it will auto assign container name within the namespace of that particular Jenkins job. PR has to be raised to correct this.

tools/rebuild.sh
pankajgoyal2: Done

arsulegai: How about `cd $TCF_HOME/common/sgx_workload || panic "Failed to change to the directory"`

Where

```
function panic()
{
    echo "[Error]: " ${*}
    exit 1
}
```

The error "No such file or directory" would be printed already. "Cannot change to directory" can occur for permissions which is not caught in the current code anyways.

manojgop: Looks good. You could rename panic to something like error_exit as the function is not existing with error code instead of panic. Also we can invoke with LINE NUM to help in debugging something like `cd $TCF_HOME/common/sgx_workload || error_exit "$LINENO: Failed to change to the directory"`

pankajgoyal2: error_exit is close to the indended change. No need to introduce $LINENO. Line No is already printed in the console when error occurs.

PR #258, opened 2020-02-07
BUILD.md: Fix `make clean` Build Instructions

danintel:
- `make clean` must run without `sudo` because it runs `pip3 uninstall`
  which must be run as non-root
- `pip3 install wheel` is required after `make clean` or build fails with
   `error: invalid command bdist_wheel`

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -136,7 +136,8 @@ The steps below will set up a Python virtual environment to run Avalon.
    cd $TCF_HOME/tools/build
    # Create virtual environment directory with name _dev
    python3 -m venv _dev
-   sudo make clean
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -136,7 +136,8 @@ The steps below will set up a Python virtual environment to run Avalon.
    cd $TCF_HOME/tools/build
    # Create virtual environment directory with name _dev
    python3 -m venv _dev
-   sudo make clean
+   make clean
+   pip3 install wheel
```
manojgop: Another option is to create an insatll target in Makefile and run this command in Makefile target and call make install

PR #259, opened 2020-02-08
PREREQUISITES.md: Fix ModuleNotFoundError with twisted and colorlog

danintel:
Currently, the following errors occur running `./tcs_startup.sh`
after a standalone build:

`ModuleNotFoundError: No module named 'twisted'`
`ModuleNotFoundError: No module named 'colorlog'`

The solution is to run this:

`pip3 install twisted colorlog`

This is despite the fact apt packages `python3-colorlog` and
`python3-twisted` are installed.

The solution was hinted at with
https://stackoverflow.com/questions/47846084/no-module-named-twisted

### DIFF
--- a/PREREQUISITES.md
+++ b/PREREQUISITES.md
@@ -89,7 +89,7 @@ sudo apt-get install -y python3-venv
 
 Also, install the following pip3 packages:
 ```bash
-pip3 install --upgrade setuptools json-rpc py-solc web3 wheel
+pip3 install --upgrade setuptools json-rpc py-solc web3 colorlog twisted wheel
 ```
 
 # <a name="docker"></a>Docker


PREREQUISITES.md
```diff
@@ -89,7 +89,7 @@ sudo apt-get install -y python3-venv
 
 Also, install the following pip3 packages:
 ```bash
-pip3 install --upgrade setuptools json-rpc py-solc web3 wheel
+pip3 install --upgrade setuptools json-rpc py-solc web3 colorlog twisted wheel
```
pankajgoyal2: colorlog, twisted and wheel packages are already mentioned as part of sudo apt-get install command. No need to install them again using pip3.

danintel: Avalon does not work unless these packages are installed with `pip3 install`. Running `apt-get install` does not work for Avalon.

TomBarnes: Those are well known python components. A BDBA scan of the executables will tell you if there are any outstanding CVEs in the version being pulled in.

PR #261, opened 2020-02-09
Move Shared KV module to top level directory.

manojgop:
- Shared KV directory is move to top level avalon directory
  from examples directory.
- Fixed lint issues.

Signed-off-by: Manoj Gopalakrishnan <manoj.gopalakrishnan@intel.com>

PR #262, opened 2020-02-10
Optimize shared kv docker image

manojgop:
- Moving shared_kv out of example and fix lint issues.
- Optimizw the docker build using multi-stage build.


Ram-srini:
Add commit description since this PR is doing multiple unrelated changes( moving shared_kv out of example, and optimizing the docker build using multi-stage build).

manojgop:
@Ram-srini This PR is dependent on https://github.com/hyperledger/avalon/pull/261. moving shared_kv out of example is implemented in https://github.com/hyperledger/avalon/pull/261

manojgop:
I'll rebase this PR after https://github.com/hyperledger/avalon/pull/261 is merged

shared_kv_storage/Dockerfile
```diff
@@ -42,26 +41,48 @@ RUN apt-get update \
...
 
-ENV TCF_HOME=/project/avalon
-
 # Make Python3 default
 RUN ln -sf /usr/bin/python3 /usr/bin/python
 
-COPY . /project/avalon/
+
+# Build Shared KV intermediate docker image.
+FROM base_image as build_image
```
pankajgoyal2: Delete the above "FROM base_image as build_image"

shared_kv_storage/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:bionic
+FROM ubuntu:bionic as base_image
```
pankajgoyal2: Change it to "FROM ubuntu:bioni as build_image"

shared_kv_storage/Dockerfile
manojgop: This is done in order to avoid including source code (eg: cpp files) in final image. We have base_image, build_image and final_image.  Final image is built based on base_image which has packages installed required to run shared_kv . Final image only copies the required build artifacts from build_image. If I remove base_image and only have build_image then final image has to be built based on build_image which will contain the source code and selectively copying required packages from build_images to run binary becomes non trivial.

pankajgoyal2: Do you have the image (avalon-lmdb-dev) size difference with this change and without this change?

shared_kv_storage/Dockerfile
```diff
@@ -42,26 +41,48 @@ RUN apt-get update \
...
 
-COPY . /project/avalon/
+
+# Build Shared KV intermediate docker image.
+FROM base_image as build_image
+
+ENV TCF_HOME=/project/avalon
+
+# Copy only required files for building shared KV build image.
+COPY VERSION /project/avalon/
```
pankajgoyal2: remove "/" at the end of above COPY command

manojgop: VERSION is a file. trailing "/" is needed. Please check the COPY command in Dockerfile

shared_kv_storage/Dockerfile
manojgop: I didn't check. I only verified that the docker container contains only required files and doesn't have files that are not required for shared kv for example contents of tc folder. But I think there should be some more room for optimizing the docker image size if we if install some packages in build_image instead of base_image

manojgop: Size reduced by 70%. 530 MB to 159 MB.

shared_kv_storage/Dockerfile
danintel: `FROM ubuntu:bioni**c** as build_image`

Ram-srini:
Please fix the commit message Optimizw -> Optimize

Issue #280, opened 2020-02-20
Make clean fails if the wheel package is already installed. Need a check to clean only if the package is available.


manojsalunke85:

danintel:
I pasted the `make clean` errors below.
A workaround is to ignore errors with `make -i clean`

```
(_dev) harrison build (master) $ make clean
/home/dano/git/tcf/tools/clean.sh
******** DELETE LMDB FILES **************
make[1]: Entering directory '/home/dano/git/tcf/common/python'
pip3 uninstall --yes dist/avalon_common-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl
Requirement 'dist/avalon_common-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl' looks like a filename, but the file does not exist
Cannot uninstall requirement avalon-common, not installed
Makefile:42: recipe for target 'clean' failed
make[1]: *** [clean] Error 1
make[1]: Leaving directory '/home/dano/git/tcf/common/python'
make[1]: Entering directory '/home/dano/git/tcf/examples/enclave_manager'
pip3 uninstall --yes dist/tcf_enclave_manager-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl
Requirement 'dist/tcf_enclave_manager-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl' looks like a filename, but the file does not exist
Cannot uninstall requirement tcf-enclave-manager, not installed
Makefile:41: recipe for target 'clean' failed
make[1]: *** [clean] Error 1
make[1]: Leaving directory '/home/dano/git/tcf/examples/enclave_manager'
(_dev) harrison build (master) $ 
```

danintel:
I fixed this with PR #281

danintel:
PR #281 integrated 2020-02-20

PR #281, opened 2020-02-20
Fix `make clean` Failure When Wheel Package is Not Installed

danintel:
Running `make clean` in `tools/build/` fails when the pip3 wheel package
is not installed with this failure:

```
pip3 uninstall --yes dist/avalon_common-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl
Requirement 'dist/avalon_common-0.5.0.dev1-cp36-cp36m-linux_x86_64.whl' looks like a filename, but the file does not exist
Cannot uninstall requirement avalon-common, not installed
Makefile:42: recipe for target 'clean' failed
```

The fix is to run `pip uninstall` only if the wheel package exists.

### DIFF
--- a/common/python/Makefile
+++ b/common/python/Makefile
@@ -32,14 +32,14 @@ build_ext :
 
 build :
 	mkdir $@
-	cd $@ 
+	cd $@
 
 install:
 	@echo INSTALLING WHEEL FILE =================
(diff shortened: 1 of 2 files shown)


danintel:
Fixes bug #280

danintel:
Merged with PR #286

PR #285, opened 2020-02-21
heart_disease_eval/README.md: Add Standalone Instructions

danintel:
- Add Standalone (non-Docker) usage instructions
- Add solcx compiler instructions

### DIFF
--- a/examples/apps/heart_disease_eval/README.md
+++ b/examples/apps/heart_disease_eval/README.md
@@ -21,22 +21,36 @@ requests on the command line.
 1.  If needed, update the Ethereum account and direct registry contract
     information in `sdk/avalon_sdk/tcf_connector.toml`
 2.  Follow instructions in the "Docker-based Build and Execution" section of
-    the [build document](../../../BUILD.md#dockerbuild) through step 5
-    (activating a virtual environment)
-3.  Terminal 1 is running the Avalon Enclave Manager and Listener with
-    `docker-compose` . Terminal 2 is running the Docker container shell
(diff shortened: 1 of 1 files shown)


examples/apps/heart_disease_eval/README.md
```diff
@@ -21,22 +21,36 @@ requests on the command line.
...
+    In that case, follow instructions in the "Standalone Build" section of
+    the [build document](../../../BUILD.md#standalonebuild)
+3.  Terminal 1 is running the Avalon Enclave Manager and Listener.
+    Terminal 2 is running either the Docker container shell (recommended) or
+    (for Standalone Builds) the Python virtual environment
 4.  In Terminal 2 run `cd $TCF_HOME/examples/apps/generic_client`
-5.  In Terminal 2, run
+5.  If you are running with Docker, then run this in Terminal 2:
+    ``` bash
+    ./generic_client.py --workload_id "heart-disease-eval" \
```
Ram-srini: This readme file is to test heart disease eval gui client right? Why do we adding generic_client commands here?

danintel: The README file shows how to run the heart disease worker many ways--from the CLI client, GUI client, and the generic_client.  In the latter case that shows how the generic_client can be used for any worker.

danintel: The generic CLI client also has the advantage of not requiring a GUI and X Windows. The Demo.py heart disease client only works with JSON files and is not interactive.

PR #291, opened 2020-02-25
Dockerfile optimizations wrt to openssl

pankajgoyal2:
Openssl is being installed and used by various Avalon components.
Installing openssl and compiling takes lot of time. This PR
addresses to create a openssl docker image which is cached in
the system and is being used by other components. This way, other
components don't have to redo the openssl fetch and compilation.

### DIFF
--- a/docker/Dockerfile.tcf-dev
+++ b/docker/Dockerfile.tcf-dev
@@ -32,6 +32,37 @@
 #   make clean && make
 
 
+# -------------=== Build openssl_image ===-------------
+
+#Build openssl intermediate docker image
+FROM ubuntu:bionic as openssl_image
+
+RUN apt-get update \
+ && apt-get install -y -q \
+    ca-certificates \
+    pkg-config \
+    make \
+    wget \
+    tar \
+ && apt-get -y -q upgrade \
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /tmp
+
+# Build ("Untrusted") OpenSSL
+RUN OPENSSL_VER=1.1.1d \
+ && wget https://www.openssl.org/source/openssl-$OPENSSL_VER.tar.gz \
+ && tar -zxf openssl-$OPENSSL_VER.tar.gz \
+ && cd openssl-$OPENSSL_VER/ \
+ && ./config \
+ && THREADS=8 \
+ && make -j$THREADS \
+ && make test \
+ && make install -j$THREADS
+
+# -------------=== build avalon image ===-------------
+#Build docker image
 FROM ubuntu:bionic
 
 # Ignore timezone prompt in apt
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.tcf-dev
```diff
@@ -111,17 +142,15 @@ RUN SGX_SDK_FILE=sgx_linux_x64_sdk_2.7.101.3.bin \
 
 
 WORKDIR /tmp
+ENV OPENSSL_VER=1.1.1d
+COPY --from=openssl_image /tmp/openssl-$OPENSSL_VER /tmp/openssl-$OPENSSL_VER/
+COPY --from=openssl_image /tmp/openssl-$OPENSSL_VER.tar.gz /tmp/
```
manojgop: We don't need to copy entire openSSL build artifacts here. copy the required openSSL build artifacts like config files, binaries and libraries ?

docker/Dockerfile.tcf-dev
```diff
@@ -111,17 +142,15 @@ RUN SGX_SDK_FILE=sgx_linux_x64_sdk_2.7.101.3.bin \
...
+
 # Build ("Untrusted") OpenSSL, then build ("trusted") SGX OpenSSL
```
manojgop: You could probably modify above comment as we are not building openSSL here ?

docker/Dockerfile.tcf-dev
pankajgoyal2: I looked at it but it creates various symlinks etc which is difficult to manage. So, its better to copy the entire openSSL build.

docker/Dockerfile.tcf-dev
pankajgoyal2: ok

docker/Dockerfile.tcf-dev
pankajgoyal2: openssl is a 3rd party tool. Currently avalon is using 1.1.1d version of it. I see that artifacts generated have version appended to them. So, we need to selectively copy the artifacts with some number as hardcoding. For eg: it generates "libcrypto.so.1.1". In future, if we decided to move to some other version of openssl, then there will be many places that we need to make change according to new version for the artifacts. So, I would prefer not to fall into that. Lets keep it like that.

manojgop: We can copy *.so or selectively copy required directories from openssl build artifacts. We can change configure command to pass build output directory. From openssl build ouput ,copy only required directories needed in run time like bin, lib etc. Please investigate further. Refer https://www.howtoforge.com/tutorial/how-to-install-openssl-from-source-on-linux/ and/or https://wiki.openssl.org/index.php/Compilation_and_Installation

pankajgoyal2: I tried copying the required directories, files and config files etc. There doesn't seem to be any improvement in docker image size.

docker/Dockerfile.tcf-dev
```diff
@@ -111,17 +142,15 @@ RUN SGX_SDK_FILE=sgx_linux_x64_sdk_2.7.101.3.bin \
...
+
+# Build ("trusted") SGX OpenSSL
+# Note: This will compile in HW or SIM mode depending on the
 # availability of /dev/isgx and /var/run/aesmd/aesm.socket
-RUN OPENSSL_VER=1.1.1d \
- && wget https://www.openssl.org/source/openssl-$OPENSSL_VER.tar.gz \
- && tar -zxf openssl-$OPENSSL_VER.tar.gz \
- && cd openssl-$OPENSSL_VER/ \
- && ./config \
+RUN cd openssl-$OPENSSL_VER/ \
```
manojgop: Does building SGXSSL using command `make SGX_MODE=${SGX_MODE} DESTDIR=/opt/intel/sgxssl all test' rebuild openSSL again ?

listener/Dockerfile
```diff
@@ -67,11 +64,35 @@ RUN OPENSSL_VER=1.1.1d \
...
+    swig \
+    software-properties-common
+
+
+ENV OPENSSL_VER=1.1.1d
+COPY --from=openssl_image /usr/local/ssl /usr/local/ssl
+COPY --from=openssl_image /usr/local/bin /usr/local/bin
+COPY --from=openssl_image /usr/local/include /usr/local/include
+COPY --from=openssl_image /usr/local/lib /usr/local/lib
+COPY --from=openssl_image /tmp/openssl-$OPENSSL_VER.tar.gz /tmp/
```
manojgop: Do we need to copy openssl-$OPENSSL_VER.tar.gz here ?

pankajgoyal2: OOPs. Its not needed here. Its copy and paste mistake. Will delete it.

docker/Dockerfile.tcf-dev
pankajgoyal2: Yes, it builds openSSL again.

PR #302, opened 2020-03-03
Optimizing and handling proxy issues with various services

pankajgoyal2:
This PR does the following:
- Handles the proxy issues with avalon-blockchain-connector service
  for fabric
- Optimizes the blockchain_connector/Dockerfile for openssl package
- Removes the various proxy settings from inside of docker containers

### DIFF
--- a/blockchain_connector/Dockerfile
+++ b/blockchain_connector/Dockerfile
@@ -21,22 +21,12 @@ ENV DEBIAN_FRONTEND=noninteractive
 # Add necessary packages
 RUN apt-get update \
  && apt-get install -y -q \
-    build-essential \
     ca-certificates \
-    pkg-config \
-    protobuf-compiler \
-    libprotobuf-dev \
-    python3-dev \
     python3-pip \
     python3-toml \
     python3-requests \
     python3-colorlog \
     python3-twisted \
-    software-properties-common \
-    dh-autoreconf \
-    ocaml \
-    xxd \
-    ocamlbuild \
  && apt-get -y -q upgrade \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
@@ -73,21 +63,45 @@ RUN OPENSSL_VER=1.1.1d \
  && make test \
  && make install -j$THREADS
 
+
 # Build Avalon blockchain connector intermediate docker image
 FROM base_image as build_image
 # Add necessary packages to build
 RUN apt-get update \
  && apt-get install -y -q \
+    pkg-config \
+    protobuf-compiler \
+    libprotobuf-dev \
+    software-properties-common \
+    dh-autoreconf \
+    ocaml \
+    xxd \
+    ocamlbuild \
     cmake \
     make \
     git \
-    wget \
-    tar \
-    unzip \
     swig \
  && apt-get -y -q upgrade \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
+
+
+# Installing web3 for ethereum
+RUN pip3 install --upgrade setuptools json-rpc web3 py-solc-x nose2 eth-abi
+
+# Install specific version of solc to support versions <0.6.0
+RUN python3 -m solcx.install v0.5.15
+
+RUN pip3 install nest_asyncio
+
+# Copy openssl build artifacts from openssl_image
+COPY --from=openssl_image /usr/local/ssl /usr/local/ssl
+COPY --from=openssl_image /usr/local/bin /usr/local/bin
+COPY --from=openssl_image /usr/local/include /usr/local/include
+COPY --from=openssl_image /usr/local/lib /usr/local/lib
+
+RUN ldconfig \
+ && ln -s /etc/ssl/certs/* /usr/local/ssl/certs/
 
 ENV TCF_HOME=/project/avalon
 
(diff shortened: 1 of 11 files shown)


docs/TestingFabricProxyModel.md
```diff
@@ -4,7 +4,15 @@
...
 2.  Set the environment variable ``TCF_HOME`` to the Avalon root directory
 
-3.  Start the Fabric network with 2 organizations, 4 peers and 3 orderers
+3.  If the host machine is behind any firewall/proxy, you need to define the
+    following parameters in /etc/environment file:
+    http_proxy=<http-proxy-url>:<port>
+    https_proxy=<https-proxy-url>:<port>
+    no_proxy=localhost,127.0.0.1,orderer3.example.com,orderer2.example.com,
+             orderer1.example.com,peer2.org1.example.com,peer1.org1.example.com,
+             peer2.org0.example.com,peer1.org0.example.com
```
manojgop: Manual approach to set no_proxy doesn't look scalable w.r.t ease of setup. Can we set these values in docker-compose file or in fabric startup script. Ideally I would like to set no_proxy only while fabric is running than set it at system level.

blockchain_connector/Dockerfile
```diff
@@ -162,3 +183,11 @@ RUN echo "TCF_HOME=/project/avalon/" >> /etc/environment \
```
manojgop: Do we also need to set no_proxy in docker container

docs/TestingFabricProxyModel.md
pankajgoyal2: Ideally, proxies should not be hard-coded in source code. This has to be set by the user in his environment. I see no-proxy has been hard-coded in scripts/start_fabric.sh for fixed number of orderers, peers and organisations. I'm going to clean this up and this has to be set by user according to the documentation provided.

manojgop: I'm not certain why should user manually set no_proxy for fabric specific configuration. I can understand setting proxies in /etc/environment manually for http_proxy or https_proxy which is company specific setting. I would let fabric startup script set it based on fabric network configuration rather than manually setting it. @litong01 Please let us know if you have any suggestions for where to set no_proxy for fabric configuration

litong01:
The proxy should be set on the machine where you host your fabric nodes. The proxy should not be set on the container. the settings should be really transparent to fabric.

blockchain_connector/Dockerfile
```diff
@@ -183,3 +182,11 @@ RUN echo "TCF_HOME=/project/avalon/" >> /etc/environment \
 
 ENV http_proxy=$http_proxy
 ENV https_proxy=$https_proxy
+ENV no_proxy=$no_proxy
```
manojgop: Do we need the lines which set http_proxy, https_proxy in lines 160-181 ?

blockchain_connector/Dockerfile
```diff
@@ -73,30 +63,36 @@ RUN OPENSSL_VER=1.1.1d \
...
-COPY ./common/python /project/avalon/common/python
-COPY ./sdk /project/avalon/sdk
-COPY ./blockchain_connector /project/avalon/blockchain_connector
+# Installing web3 for ethereum
+RUN pip3 install --upgrade setuptools json-rpc web3 py-solc-x nose2 eth-abi
+
+# Install specific version of solc to support versions <0.6.0
+RUN python3 -m solcx.install v0.5.15
+
+RUN pip3 install nest_asyncio
```
manojgop: pip packages like json-rpc web3 py-solc-x nose2 eth-abi is also needed at run time and should ideally be installed in base image. I guess things are working now because of below line in final image
COPY --from=build_image /usr/local/ /usr/local. This approach also is not optimal. We should be copy only required packages in final image instead of copying entire /usr/local

enclave_manager/Dockerfile
```diff
@@ -160,8 +160,6 @@ ARG MAKECLEAN
 ARG TCF_DEBUG_BUILD
 
 # Environment setup
-ENV http_proxy=$http_proxy
-ENV https_proxy=$https_proxy
```
manojgop: Have we tested SGX h/w with this change. Please verify if connection to IAS server works after removing this.

enclave_manager/Dockerfile
```diff
@@ -232,8 +230,6 @@ RUN echo "Building avalon enclave manager" \
 # Build Final image and install dependent modules
 FROM base_image as final_image
 
-ENV http_proxy=$http_proxy
-ENV https_proxy=$https_proxy
```
manojgop: Have we tested SGX h/w with this change. Please verify if connection to IAS server works after removing this.

blockchain_connector/Dockerfile
pankajgoyal2: To be handled in a seperate PR. Issue raised https://github.com/hyperledger/avalon/issues/307 to track this.

pankajgoyal2:
> Since build is failing with proxy settings. Please fix that.

already fixed and pushed the changes. @Ram-srini @manojgop review please

docker-compose-sgx.yaml
```diff
@@ -22,10 +22,14 @@ services:
       context: .
       dockerfile: ./docker/Dockerfile
       args:
+        - DISPLAY=${DISPLAY:-}
```
Ram-srini: DISPLAY is required in run time not for build. Please move it to environment.

docker-compose.yaml
```diff
@@ -22,11 +22,15 @@ services:
       context: .
       dockerfile: ./docker/Dockerfile
       args:
+        - DISPLAY=${DISPLAY:-}
+        - XAUTHORITY=~/.Xauthority
```
Ram-srini: XAUTHORITY is also required during run time, move it to environment section.

docker/Dockerfile
```diff
@@ -161,21 +153,19 @@ ARG DISPLAY
 ARG XAUTHORITY
 
 #Environment setup
-ENV http_proxy=$http_proxy
-ENV https_proxy=$https_proxy
```
manojgop: if proxy is not set inside container image, I assume we need to find a different way to enable proxy for kubernetes setup which doesn't use docker-compose.

PR #303, opened 2020-03-04
Fix RSA crypro error for Besu

rranjan3:
Check Besu blockchain for worker before adding. If worker
already present, make a workerUpdate contract call.
Worker details were not being updated if the same network
of Besu was being used across restarts of Avalon. This caused
stale verification/encryption key to be used in consequent runs
after a worker is added to a fresh blockchain.

### DIFF
--- a/blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
+++ b/blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
@@ -43,6 +43,9 @@ logging.basicConfig(
     format="%(asctime)s - %(levelname)s - %(message)s", level=logging.INFO)
 
 LISTENER_SLEEP_DURATION = 5  # seconds
+# Return codes
+SUCCESS = 0
+ERROR = 1
 
 
 class EthereumConnector:
@@ -70,6 +73,8 @@ class EthereumConnector:
             .get_contract_instance(
                 worker_reg_contract_file, worker_reg_contract_address)
 
+        self._worker_registry = EthereumWorkerRegistryImpl(config)
+
         work_order_contract_file = tcf_home + "/" + \
             config["ethereum"]["work_order_contract_file"]
         work_order_contract_address = \
@@ -79,7 +84,7 @@ class EthereumConnector:
             .get_contract_instance(
                 work_order_contract_file, work_order_contract_address)
 
-        self._worker_registry = EthereumWorkerRegistryImpl(config)
+        self._work_order_proxy = EthereumWorkOrderProxyImpl(config)
 
     def _lookup_workers_in_kv_storage(self, worker_registry):
         """
@@ -105,7 +110,8 @@ class EthereumConnector:
             logging.error("Failed to lookup worker")
         return []
 
-    def _retrieve_worker_details_from_kv_storage(self, worker_registry, worker_id):
+    def _retrieve_worker_details_from_kv_storage(self, worker_registry,
+                                                 worker_id):
         # Retrieve worker details
         jrpc_req_id = random.randint(0, 100000)
         worker_info = worker_registry.worker_retrieve(worker_id, jrpc_req_id)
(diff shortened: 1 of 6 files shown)


blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -120,27 +123,58 @@ def _retrieve_first_worker_details(self):
...
-            logging.error("Error while adding worker to ethereum"
+            logging.error("Error while adding/updating worker to ethereum"
                           + " blockchain : "+str(e))
 
+    def _lookup_workers_on_chain(self):
+        """
+        Lookup all workers on chain to sync up with kv storage
+        """
+        jrpc_req_id = random.randint(0, 100000)
+        # TODO: Remove hardcoding and pass wild characters instead
```
Ram-srini: Are we deleting old workers after restarting avalon?

rranjan3: No @Ram-srini. The old worker is not being deleted. When Avalon(connector) restarts, it tries to see what all workers already exist on the chain. If there's a worker already, it's details are only updated.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -120,27 +123,58 @@ def _retrieve_first_worker_details(self):
...
+            txn_dict = self._worker_reg_contract_instance.functions\
+                .workerUpdate(worker_id, details)\
+                .buildTransaction(self._eth_client.get_transaction_params())
+        else:
+            logging.info("Adding new worker {} to ethereum blockchain"\
+                .format(worker_id_hex))
+            txn_dict = self._worker_reg_contract_instance.functions\
+                .workerRegister(worker_id, worker_type, org_id,
+                                [app_type_id], details)\
+                .buildTransaction(self._eth_client.get_transaction_params())
```
Ram-srini: Can we move this buildTransaction to workerRegister()?

rranjan3: @Ram-srini, I will accommodate this in another PR. Replacing these with sdk calls would be the way to go.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -98,48 +98,94 @@ def _retrieve_first_worker_details(self):
...
+            app_type_id = self._eth_client.get_bytes_from_hex(
+                worker_info["applicationTypeId"])
+            details = json.dumps(worker_info["details"])
+
+            if wid in wids_onchain:
+                logging.info("Updating worker {} on ethereum blockchain"\
+                    .format(wid))
+                txn_dict = self._worker_reg_contract_instance.functions\
+                    .workerUpdate(worker_id, details)\
+                    .buildTransaction(self._eth_client.get_transaction_params())
```
manojgop: Can we use SDK functions to call ethereum contract

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -98,48 +98,94 @@ def _retrieve_first_worker_details(self):
...
+                new_workers_kv.append(wid)
+            try:
+                txn_receipt = self._eth_client.execute_transaction(txn_dict)
+            except Exception as e:
+                logging.error("Error while adding/updating worker to ethereum"
+                              + " blockchain : "+str(e))
+        for wid in wids_onchain:
+            # Mark all stale workers on blockchain as decommissioned
+            if wid not in new_workers_kv:
```
manojgop: if a worker id is present in shared kv and on-chain then onchain worker gets updated.  All active workers present in shared KV and on-chain are valid. Only those workers which are not present in shared KV can be made stale. Looks like above logic only considers new workers in shared KV.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -98,48 +98,94 @@ def _retrieve_first_worker_details(self):
...
+    def _lookup_workers_onchain(self):
+        """
+        Lookup all workers on chain to sync up with kv storage
+        """
+        jrpc_req_id = random.randint(0, 100000)
+        # TODO: Remove hardcoding and pass wild characters instead
+        worker_lookup_result = self._worker_registry.worker_lookup(
+            WorkerType.TEE_SGX,
+            'aabbcc1234ddeeff',
+            '11aa22bb33cc44dd',
```
manojgop: Read these values from config file

examples/apps/generic_client/eth_generic_client.py
```diff
@@ -419,34 +391,53 @@ def _handle_ethereum_event(event, *kargs, **kwargs):
...
+    this data to get the first worker that has Active status.
+    """
+    jrpc_req_id = random.randint(0, 100000)
+    worker_retrieve_result = None
+    if not worker_id:
+        # Lookup all worker id present on  blockchain
+        worker_lookup_result = worker_registry.worker_lookup(
+            WorkerType.TEE_SGX,
+            'aabbcc1234ddeeff',
+            '11aa22bb33cc44dd',
```
manojgop: read values from config file. or you could match any org/app ids

examples/apps/generic_client/eth_generic_client.py
```diff
@@ -546,16 +537,9 @@ def Main(args=None):
...
     worker_registry = _create_worker_registry_instance(blockchain, config)
-    if not worker_id:
-        # Get first worker from worker registry
-        worker_id = _lookup_first_worker(
-            blockchain, worker_registry, jrpc_req_id)
-        if worker_id is None:
-            logger.error("\n Unable to get worker \n")
-            sys.exit(-1)
+    worker_obj, worker_id = _get_first_active_worker(worker_registry,
+                                                    worker_id)
```
manojgop: In case of error, return None. Check return value and exit in case of error.

examples/apps/generic_client/eth_generic_client.py
```diff
@@ -581,6 +564,7 @@ def Main(args=None):
             exit(1)
 
     # Submit work order
+    jrpc_req_id = random.randint(0, 100000)
```
manojgop: I think we can use either this approach or follow existing approach (line 571) which increments jrpc_req_id for each request. We don't need both.

dcmiddle: What is this field used for? If this should be a unique ID then I recommend using UUID.

rranjan3: Good suggestion @dcmiddle . This would be the way to go. 

It is a [field](https://www.jsonrpc.org/specification#request_object) used to verify the context of a response receive after posting a JRPC request. As of now, it is not being verified in [Avalon's sdk](https://github.com/hyperledger/avalon/blob/master/sdk/avalon_sdk/http_client/http_jrpc_client.py#L56) as the communication is over http and only a single call is included in each call.

When there is significant traffic, multi-call JRPC requests might be possible in a single HTTP request. This field would play a role there to map requests to responses. A shift from HTTP(synchronous request-response) would also require proper handling of this field.

danintel: Added to FAQ. PR #309

examples/apps/generic_client/eth_generic_client.py
rranjan3: @manojgop, raised an [issue](https://github.com/hyperledger/avalon/issues/310) for this. Will remove from elsewhere as well.

PR #308, opened 2020-03-06
Generate separate python package for crypto module

manju956:
Ths isolation of crypto package avoids installation of crypto module
by any module which depends on just common/python but not crypto.


docker/Dockerfile
```diff
@@ -143,6 +144,11 @@ WORKDIR /project/avalon/common/python
 RUN echo "Building Avalon Common Python\n" \
   && make
 
+WORKDIR /project/avalon/common/crypto
+
+RUN echo "Building Avalon Common Crypto Python\n" \
+  && make
```
manojgop: @Ram-srini @rranjan3 . After merging this PR, will require related changes in block chain connector module to remove crypto dependencies. Can be done as separate PR

listener/Dockerfile
```diff
@@ -140,6 +142,22 @@ ENV TCF_HOME=/project/avalon
...
+COPY --from=build_image /project/avalon/common/python/dist/*.whl dist/
+COPY --from=build_image /project/avalon/common/crypto/dist/*.whl dist/
+COPY --from=build_image /project/avalon/sdk/dist/*.whl dist/
+COPY --from=build_image /project/avalon/listener/dist/*.whl dist/
+
+# Installing wheel file requires python3-pip package.
+# But python3-pip package will increase size of final docker image.
+# So remove python3-pip package and dependencies after installing wheel file.
+RUN apt-get update \
+ && apt-get install -y -q python3-pip \
```
pankajgoyal2: python3-pip package is already part of base_image. No need to install it again here.

manojgop: python json-rpc package may not be needed during build stage. This could be installed in final image instead of base image. so python-pip3 is not required to be installed in base image.

manju956: Done

manju956:
> you are including verify_report as part of crypto package. How is that related?

Yes, because verify_report module also does crypto operations like signature verification, cert chain verification etc.

pankajgoyal2:
> you are including verify_report as part of crypto package. How is that related?

With this PR you are intending to have the following directory structure:
-> common
          |-> crypto
                     -> crypto_utils
                                  |-> crypto
                                  |-> verify_report
                    -> Makefile
                    -> setup.py

I understand the objective of this PR is to avoid other avalon components to have crypto packages installed if they don't need them.
My recommendation is as follows:
1. Either don't keep verify_report as part of crypto folder 
      OR
2. Rename top level crypto and crytp_utils to some other name like <abc> and <abc>_utils and inside <abc>_utils we have 2 directories: crypto (for crypto operations) and verify_report (for IAS report verification).

manojgop:
> > you are including verify_report as part of crypto package. How is that related?
> 
> With this PR you are intending to have the following directory structure:
> -> common
> |-> crypto
> -> crypto_utils
> |-> crypto
> |-> verify_report
> -> Makefile
> -> setup.py
> 
> I understand the objective of this PR is to avoid other avalon components to have crypto packages installed if they don't need them.
> My recommendation is as follows:
> 
> 1. Either don't keep verify_report as part of crypto folder
>    OR
> 2. Rename top level crypto and crytp_utils to some other name like  and _utils and inside _utils we have 2 directories: crypto (for crypto operations) and verify_report (for IAS report verification).

We can rename as below. These are basically crypto utilities for Avalon
common/crypto -> common/crypto_utils 
common/crypto/crypto_utils -> common/crypto_utils/avalon_crypto_utils

pankajgoyal2:
I don't want crypto as part of top level directory name if verify_report is part of this package.

manojgop:
> I don't want crypto as part of top level directory name if verify_report is part of this package.

verify report is also doing crypto operations. So we can rename the package to avlaon crypto utils.

pankajgoyal2:
> > I don't want crypto as part of top level directory name if verify_report is part of this package.
> 
> verify report is also doing crypto operations. So we can rename the package to avlaon crypto utils.

Ok

common/crypto_utils/setup.py
```diff
@@ -0,0 +1,121 @@
...
+    verify_report_modulefiles,
+    swig_opts=['-c++'] + ['-I%s' % i for i in verify_report_include_dirs],
+    extra_compile_args=compile_args,
+    include_dirs=verify_report_include_dirs,
+    library_dirs=library_dirs,
+    libraries=libraries)
+
+# -----------------------------------------------------------------
+setup(
+    name='avalon_crypto',
```
pankajgoyal2: change it to avalon_crypto_utils

common/crypto_utils/setup.py
```diff
@@ -0,0 +1,121 @@
...
+    version=version,
+    description='Avalon Crypto Library',
```
pankajgoyal2: 'Avalon Crypto Utils Library'

tools/rebuild.sh
```diff
@@ -186,12 +186,18 @@ cd build
...
 try make "-j$NUM_CORES"
 
-yell --------------- EXAMPLES COMMON PYTHON ---------------
+yell --------------- COMMON PYTHON ---------------
 cd $TCF_HOME/common/python || error_exit "Failed to change to the directory"
 
 try make "-j$NUM_CORES"
 try make install
 
+yell --------------- COMMON CRYPTO PYTHON ---------------
```
pankajgoyal2: COMMON CRYPTO UTILS

common/crypto_utils/Makefile
```diff
@@ -0,0 +1,51 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
+MOD_VERSION=${shell ../../bin/get_version}
+
+# Format of wheel package is <pkg_name>-{python-tag}-{abi-tag}-{platform}
+WHEEL_FILE=dist/avalon_crypto-${MOD_VERSION}-cp${PY_VERSION}-cp${PY_VERSION}m-linux_x86_64.whl
```
pankajgoyal2: change it to dist/avalon_crypto_utils-${MOD_VERSION}.....

common/crypto_utils/setup.py
manju956: Done

common/crypto_utils/setup.py
manju956: Done

tools/rebuild.sh
manju956: Done

common/crypto_utils/Makefile
manju956: Done

PR #314, opened 2020-03-10
Fix enclave security issues

manju956:
This commit address followin security concerns
- Sanitization of local variables on stack containing secrets
- Sanitization of local variables before reassigning them new values
- Signed type used for indexing or size are converted to unsigned
- Defined destructors to sanitize objects from heap memory

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
@@ -19,17 +19,25 @@
 #include <stdlib.h>
 #include <string>
 #include <vector>
+#include <string.h>
+#include<mbusafecrt.h>
 
 #include "crypto.h"
 #include "error.h"
 #include "avalon_sgx_error.h"
 #include "tcf_error.h"
 #include "zero.h"
+#include "enclave_utils.h"
 
 #include "jsonvalue.h"
 #include "parson.h"
 
 #include "enclave_data.h"
+
+// Below constants are the max limit for serialized version(base64 encoded)
+// of private signature and private encryption keys
+#define MAX_SERIALIZED_SIG_PRIV_KEY_SIZE 256
+#define MAX_SERIALIZED_ENC_PRIV_KEY_SIZE 2048
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
@@ -55,6 +63,15 @@ EnclaveData::EnclaveData(void) {
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+EnclaveData::~EnclaveData(void) {
+    // Sanitizing class instance members other than below leads to enclave
+    // crash hence avoid the sanitization.
+    serialized_private_data_.clear();
+    serialized_public_data_.clear();
+}
+
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 EnclaveData::EnclaveData(const uint8_t* inSealedData) {
     tcf::error::ThrowIfNull(inSealedData, "Sealed sign up data pointer is NULL");
 
@@ -74,6 +91,10 @@ EnclaveData::EnclaveData(const uint8_t* inSealedData) {
     std::string decrypted_data_string(reinterpret_cast<const char*>(&decrypted_data[0]));
     DeserializeSealedData(decrypted_data_string);
 
+    // Clear local variable storing secret(s)
+    decrypted_data.clear();
+    decrypted_data_string.clear();
+
     SerializePrivateData();
     SerializePublicData();
 }
@@ -83,6 +104,7 @@ EnclaveData::EnclaveData(const uint8_t* inSealedData) {
 void EnclaveData::DeserializeSealedData(const std::string& inSerializedEnclaveData) {
     std::string svalue;
     const char* pvalue = nullptr;
+    size_t pvalue_len = 0;
 
     // Parse the incoming wait certificate
     JsonValue parsed(json_parse_string(inSerializedEnclaveData.c_str()));
@@ -95,7 +117,6 @@ void EnclaveData::DeserializeSealedData(const std::string& inSerializedEnclaveDa
     pvalue = json_object_dotget_string(keystore_object, "SigningKey.PublicKey");
     tcf::error::ThrowIf<tcf::error::ValueError>(
         !pvalue, "Failed to retrieve public signing key from the key store");
-
     svalue.assign(pvalue);
     public_signing_key_.Deserialize(svalue);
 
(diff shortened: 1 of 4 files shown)


tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -74,6 +81,10 @@ EnclaveData::EnclaveData(const uint8_t* inSealedData) {
     std::string decrypted_data_string(reinterpret_cast<const char*>(&decrypted_data[0]));
     DeserializeSealedData(decrypted_data_string);
 
+    // clear local varaible storing secret(s)
```
manojgop: varaible  > variable

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -99,6 +110,10 @@ void EnclaveData::DeserializeSealedData(const std::string& inSerializedEnclaveDa
     svalue.assign(pvalue);
     public_signing_key_.Deserialize(svalue);
 
+    // Clear local variables storing secrets before re-assigning new value
+    svalue.clear();
+    std::string(pvalue).clear();
```
manojgop: Directly clear the contents of pvalue ( memset_s). std:string() constructor from char* does a copy.

tc/sgx/trusted_worker_manager/enclave/work_order_processor.h
```diff
@@ -27,6 +27,11 @@ namespace tcf {
         class WorkOrderProcessor {
         public:
                 WorkOrderProcessor() {}
+                ~WorkOrderProcessor() {
+                    // Sanitize class members storing secrets
+                    worker_encryption_key.clear();
```
Ram-srini: Are these class instance variables?

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -469,8 +469,8 @@ namespace tcf {
                     "Signature verification of client request failed. Request is tampered.");
             }
             std::vector<tcf::WorkOrderData> wo_data = ExecuteWorkOrder();
-            int i = 0;
-            int out_data_size = data_items_out.size();
+            size_t i = 0;
+            size_t out_data_size = data_items_out.size();
```
Ram-srini: Change in other places in the same file.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -53,6 +53,13 @@ EnclaveData::EnclaveData(void) {
     SerializePublicData();
 }
 
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+EnclaveData::~EnclaveData(void) {
+    serialized_private_data_.clear();
+    serialized_public_data_.clear();
```
Ram-srini: Why only these variable, class has other instance variables are also.

tc/sgx/trusted_worker_manager/enclave/work_order_processor.h
manju956: that's correct. i am clearing the variables whose values are sensitive and need to be confidential.

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
manju956: Will do.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: Ok

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: Ok

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: I tried calling destructors of other class members which are inturn objects of different class (related to crypto), but enclave seems to be crashing. Hence i tried not calling destructor explicitly.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -93,35 +107,55 @@ void EnclaveData::DeserializeSealedData(const std::string& inSerializedEnclaveDa
 
     // Public signing key
     pvalue = json_object_dotget_string(keystore_object, "SigningKey.PublicKey");
+    pvalue_len = strlen(pvalue);
```
manojgop: Do this operation after ValueError check for null pointer. Try to use strnlen_s and max szie  to check will be size of public key,

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -93,35 +107,55 @@ void EnclaveData::DeserializeSealedData(const std::string& inSerializedEnclaveDa
...
     tcf::error::ThrowIf<tcf::error::ValueError>(
         !pvalue, "Failed to retrieve public signing key from the key store");
 
     svalue.assign(pvalue);
     public_signing_key_.Deserialize(svalue);
 
+    // Clear local variables storing secrets before re-assigning new value
+    svalue.clear();
+    memset_s((char*)pvalue, pvalue_len, 0, pvalue_len);
```
manojgop: Max size to set in memset_s will be size of key

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: I can add strlen check after value error validation. Currently we don't have max size of public key. We have max size of public enclave data. So, i am not sure how to strlen_s function here.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: We don't have max size of key like public signing key or public encryption key.

manojgop: Use Max expected size of public key. memset_s will check if pvalue_len < MAX_SIZE and will not result in any buffer overflow while copying.

manju956: I understand but in Avalon we haven't defined max size of public key.

danintel: FYI, in the future, with quantum-computing resistant crypto, the public key may be quite large.

manojgop: @manju956 define a MAX_KEY_SIZE based on your current implementation and check that buffer size is less than MAX_SIZE.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
danintel: Please add a comment about that in the code.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: I could compute strlen upto size of serialized encryption/signing keys as the keys stored in these local variables are infact serialized version.

manju956: these are serialized version of keys, so i am not sure how to define MAX_KEY_SIZE for different keys.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
manju956: Ok

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -19,18 +19,26 @@
...
 #include "tcf_error.h"
 #include "zero.h"
+#include "enclave_utils.h"
 
 #include "jsonvalue.h"
 #include "parson.h"
 
 #include "enclave_data.h"
 
+// Below constants are the max limit for serialized version(base64 encoded)
```
Ram-srini: Good coding practice is to keep these constants in header file.

manju956: Ok, will move it to header file.

manojgop: It depends. If these constants/pre-processor symbols has be shared across multiple source files then keeping them in header file is suggested. But if these are to be used only locally then keeping them in source files is the better approach.

manju956: This constant is only used by enclave_data.cpp to calculate string length safely.

PR #316, opened 2020-03-11
Update Kubernetes deployment files.

manojgop:
- Update K8s yaml files to use Avalon containers for
  shared kv, enclave manager, listener and shell.

- Copy scripts in Avalon shell container


docker/Dockerfile
```diff
@@ -182,6 +182,9 @@ COPY --from=build_image /project/avalon/common/python/dist/*.whl dist/
 COPY --from=build_image /project/avalon/common/crypto_utils/dist/*.whl dist/
 COPY --from=build_image /project/avalon/sdk/dist/*.whl dist/
 
+# Copy Scripts
+COPY ./scripts /project/avalon/scripts
```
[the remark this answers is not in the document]
manojgop: We need scripts since apps run in avalon-shell container and thiis script invokes generic client application. There will be more scripts added in this directory. So for now copying the entire scripts folder in avalon-shell container. Later if we decide to create separate containers for applications we can revisit this.

docker/k8s/listener.yaml
```diff
@@ -0,0 +1,32 @@
...
+      name: avalon-listener
+      labels:
+        app: avalon-listener
+    spec:
+      containers:
+      - image: avalon-listener-dev:latest
+        name: avalon-listener
+        imagePullPolicy: Never
+        command: ["/bin/bash"]
+        args: ["-c", "avalon_listener --bind http://0.0.0.0:8080 --lmdb_url http://avalon-lmdb:9090"]
```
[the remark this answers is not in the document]
manojgop: We can use any port. Good thing is ports other than 1947 (which is added as default in config file) also works in K8s environment. Do you suggest to use 1947 for any specific reason ?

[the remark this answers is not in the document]
danintel: It makes it easier to debug and understand if we use the same "well known port number", which for the Avalon listener is TCP port 1947.

docker/k8s/README.md
```diff
@@ -36,21 +28,26 @@ in SGX-SIM mode.
...
-5.  Start `TCS Listener` (to accept and handle requests from clients)
+3.  Start `http jrpc Listener service` (to accept and handle http jrpc requests from clients)
     ```bash
-    kubectl create -f tcs.yaml
+    kubectl create -f listener.yaml
     ```
 
 ## Propose a request
-1. Send the request 
+1. Before sending a workload request check if sharedkv, enclave manager and listerner pods are running
```
danintel: Change listerner to listener.

docker/k8s/listener.yaml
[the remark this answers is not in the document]
danintel: It makes it easier to debug and understand if we use the same "well known port number", which for the Avalon listener is TCP port 1947.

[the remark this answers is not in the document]
manojgop: Done

docker/k8s/README.md
manojgop: Done

docker/Dockerfile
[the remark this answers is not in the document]
Ram-srini: Why can't we generic client inside the shell container? As we are running in docker.

docker/k8s/sgx-enclave-manager-deployment.yaml
```diff
@@ -16,21 +16,21 @@ version: '3.5'
...
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: tee
+      app: enclave-manager-sgx
   template:
     metadata:
-      name: tee
+      name: enclave-manager-sgx
```
Ram-srini: Can we use the naming convention same as in docker, change to  enclave-manager-sgx to enclave-manager. So that there is consistency throughout.

docker/k8s/README.md
```diff
@@ -36,21 +28,26 @@ in SGX-SIM mode.
...
+2. Send the request 
     ```bash
     kubectl create -f propose-requests-job.yaml
     ```
 
     > The above command would run the 
     > [propose_requests.sh](./scripts/propose_requests.sh) as a kubernetes job
 
-2. Check the request handling logs (the request should be done in several 
+3. Check the request handling logs (the request should be done in several
```
Ram-srini: How do we run other workload for example echo using generic client? Getting into avalon shell and running generic command would be nicer approach..

docker/Dockerfile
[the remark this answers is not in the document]
pankajgoyal2: I think copying scripts is not needed. This kubernetes setup is to let people try, setup and run whatever they want. Anyone can go to shell container and run whatever they want.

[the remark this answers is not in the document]
manojgop: @pankajgoyal2 The scripts are copied since kubernetes job pod uses it to start the job(generic_client) when pods are launched. Please suggest how to run kubernetes job pod without copying scripts to avalon shell container.

docker/k8s/README.md
manojgop: This is an example kubernetes job pod to create and run a workloads automatically without user intervention. In this case the job pod runs heart-disease workload.  If we use kubernetes pod to create a shell then user has to manually get into the pod container. We can do this as a separate Kubernetes deployment file in a different PR. But purpose of job pods is is to runs jobs (workloads) to completion.  https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/.

docker/k8s/sgx-enclave-manager-deployment.yaml
manojgop: This is in order to distinguish enclave manger container with and without SGX support.  I just noticed that the naming convention for enclave manger docker file with SGX support need to change. The docker image  for enclave manager with and without SGX support should be different. @manju956 could you please take care of it

docker/k8s/README.md
```diff
@@ -61,8 +58,16 @@ in SGX-SIM mode.
...
+    ```bash
+    kubectl create -f shell.yaml
+    ```
+2. After verifying avalon shell pod is running get a shell to the running container
+    ```bash
+    # Let's say 'avalon-shell-abcde' is the pod created above
+    kubectl exec -it avalon-shell-abcde -- /bin/bash
+    ```
+
+# Heads up
```
Ram-srini: Please add steps to run the generic_client command

manojgop: This is not related to running any specific application. Shell can be used to run any application manually. I've mentioned in in step 1.

docker/k8s/lmdb.yaml
```diff
@@ -6,23 +6,19 @@ spec:
...
+      - image: avalon-lmdb-dev:latest
         name: lmdb-server
         imagePullPolicy: Never
-        envFrom:
-        - prefix: TCF_
-          configMapRef:
-            name: tcf-config
         command: ["/bin/bash"]
-        args: ["$(TCF_HOME)/scripts/lmdb.sh"]
+        args: ["-c", "kv_storage --bind http://0.0.0.0:9090"]
```
pankajgoyal2: why do you use 0.0.0.0 instead of localhost?

manojgop: This is to bind to any IPv4 address which is not allocated yet. Container pods IP will be dynamically allocated by K8s. But other pods will communicate to it via service abstraction exposed by this pod

PR #320, opened 2020-03-13
Handle race for worker sync-up in Ethereum proxy

rranjan3:
- Add dependency of enclave-manager on lmdb in docker-compose
- Purge stale connectors directory in common/python


blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -41,7 +42,7 @@
 logging.basicConfig(
     format="%(asctime)s - %(levelname)s - %(message)s", level=logging.INFO)
 
-LISTENER_SLEEP_DURATION = 5  # seconds
+WORKER_SYNCUP_SLEEP_DURATION = 5  # seconds
```
manju956: Can we put this duration in config file?

rranjan3: It's more of an implementation construct. Hence it's not worth exposing it to be configured from outside.

Ram-srini: Why do we need this sleep duration? Moment we start listener it will have workers available in lmdb because enclave adds to that worker entry. Add a dependency in docker to start connector after enclave manager and listener.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -104,7 +105,13 @@ def _lookup_workers_in_kv_storage(self, worker_registry):
             if worker_lookup_result["result"]["totalCount"] != 0:
                 return worker_lookup_result["result"]["ids"]
             else:
-                logging.error("No workers found")
+                logging.error("No workers found in kv storage")
+                # Make recursive calls upto 3 times if no workers are found 
+                if attempt_counter < 3:
+                    # Sleep for 2 seconds before next call
+                    time.sleep(WORKER_SYNCUP_SLEEP_DURATION)
```
Ram-srini: In fabric this is not case, because of dependency in the containers.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
rranjan3: There's [dependency in docker-compose](https://github.com/hyperledger/avalon/blob/master/docker-compose-eth-besu.yaml#L140) already. But, I saw this happen a couple of times when I was repetitively doing docker-compose down & up. Hence this change.

manju956: makes sense.

blockchain_connector/avalon_blockchain_connector/ethereum/ethereum_connector.py
```diff
@@ -104,7 +105,13 @@ def _lookup_workers_in_kv_storage(self, worker_registry):
...
+                    self._lookup_workers_in_kv_storage(worker_registry,
+                                                       attempt_counter+1)
```
manojgop: Is this tested with recursive calls. IMO we should assign the return value to a variable and return this at end of this function. Else you can call this function from caller in a loop in a non-recursive fashion.
```
if attempt_counter < 3:
    # Sleep for 2 seconds before next call
    time.sleep(WORKER_SYNCUP_SLEEP_DURATION)
    worker_list = self._lookup_workers_in_kv_storage(worker_registry,
                                                       attempt_counter+1)
else 
     logging.error("Failed to lookup worker")
     return []
return worker_list
```

rranjan3: Good suggestion @manojgop . Recursion is not something easy to follow at times. Though this would not be required anymore.
@manojgop @Ram-srini @manju956 Dependency of enclave-manager on lmdb was what went missing. Revisiting the startup logs helped. Was not consistently  reproducable which is why could not verify earlier.

PR #324, opened 2020-03-13
Add Makefile for `shared_kv_storage/db_store`

danintel:
Replaces manually running mkdir, cmake, and make.


manojgop:
Nice @danintel . This looks much cleaner. We need to follow this approach in other modules too.
For example, build dependent CPP modules as part enclave manager Makefile.

PR #326, opened 2020-03-16
Fix fabric startup script issue

Ram-srini:
New minifab by default won't generate the network profiles.
Added command to generate the network profiles.


docker-compose-fabric.yaml
```diff
@@ -36,7 +36,7 @@ services:
       - ./examples/apps/generic_client/:/project/avalon/examples/apps/generic_client/
       - ~/.Xauthority:/root/.Xauthority
       - ~/mywork/vars/keyfiles:/keyfiles
-      - ~/mywork/vars/mychannel_network.json:/project/avalon/sdk/avalon_sdk/fabric/network.json
+      - ~/mywork/vars/profiles/mychannel_network_for_pysdk.json:/project/avalon/sdk/avalon_sdk/fabric/network.json
```
manojgop: How is the file "mychannel_network_for_pysdk.json" getting created ?

scripts/start_fabric.sh
```diff
@@ -83,7 +83,10 @@ fi
 
 # If already fabric network up and running skip start again.
 if [[ ! $(docker ps --format '{{.Names}}' |grep "peer*") ]]; then
+    # Start the fabric containers peers, orderers, cas
     minifab up -i $FABRIC_VERSION
+    # Create network profile file for avalon to use
+    minifab profilegen
```
manojgop: Please add a comment which file is created by execution of this command ?

docker-compose-fabric.yaml
Ram-srini: It is generated from minifab in work_dir(~/mywork/var

scripts/start_fabric.sh
Ram-srini: yes

docker-compose-fabric.yaml
manojgop: Please add a comment in docker-compose file and/or startup script

scripts/start_fabric.sh
manojgop: Is there an option to pass generated file name with minifab profilegen ? If not, please add a comment which file this command will create

PR #332, opened 2020-03-18
Fix SGX HW mode

manju956:
Avalon workflow is broken in SGX HW mode.
Following SGX error is observed during enclave initialization.
failed to initialize enclave; Failed to get SGX quote size.: UNKNOWN SGX ERROR:<error_value>

This issue is observed since changes to containerizing enclave manager was introduced.
Also, the issue is due to some discrepency in final image of enclave manager dockerfile.
As a temporary workaround, final image is removed. All the necessary binaries and environement
will be available in build image of dockerfile itself.

### DIFF
--- a/enclave_manager/Dockerfile
+++ b/enclave_manager/Dockerfile
@@ -218,51 +218,26 @@ RUN mkdir -p build \
 
 WORKDIR /project/avalon/common/python
 
-RUN make
+RUN make && make install
 
 WORKDIR /project/avalon/common/crypto_utils
 
-RUN make
+RUN make && make install
 
 WORKDIR /project/avalon/sdk
 
-RUN make
+RUN make && make install
 
 WORKDIR /project/avalon/enclave_manager
 
-RUN echo "Building avalon enclave manager" \
-  && make
+RUN echo "Build and Install avalon enclave manager" \
+  && make && make install
 
-# Build Final image and install dependent modules
-FROM base_image as final_image
+FROM build_image as final_image
 
-ENV TCF_HOME=/project/avalon
-
-WORKDIR /project/avalon/
-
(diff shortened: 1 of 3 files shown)


pankajgoyal2:
I'm approving this PR but raise an issue in github to optimize enclave_manager/Dockerfile so as to reduce the final image size!

manju956:
> I'm approving this PR but raise an issue in github to optimize enclave_manager/Dockerfile so as to reduce the final image size!

Raised git issue https://github.com/hyperledger/avalon/issues/333

enclave_manager/Dockerfile
```diff
@@ -218,51 +218,26 @@ RUN mkdir -p build \
...
-# So remove python3-pip package and dependencies after installing wheel file.
-RUN apt-get update \
- && apt-get install -y -q python3-pip \
- && echo "Install Common Python, SDK and Enclave Manager packages\n" \
- && pip3 install dist/*.whl \
- && echo "Remove unused packages from image\n" \
+# python3-pip package will increase size of final docker image.
+# So remove python3-pip package and dependencies
+RUN echo "Remove unused packages from image\n" \
  && apt-get autoremove --purge -y -q python3-pip \
```
arsulegai: Maybe there are more packages that are required during build only? (For example, please confirm if `ocamlbuild` is required, I see this also is installed in build_image)

manju956: True, thats the reason we use to copy required packages from build stage to final stage earlier. But we have hit some blocker issue in HW mode which is due to optimization done in final stage.
Please note this is a temporary enable of SGX HW mode. We still need to fix the issue and bring back the optimization.
This can be tracked in git issue - https://github.com/hyperledger/avalon/issues/333

arsulegai: Ok

PR #337, opened 2020-03-18
Add Doxygen Support

danintel:
Add Makefile and Doxyfile to run Doxygen,
the automatic document from source generator.

For this first phase, only documentation for Inside-Out I/O is generated.
Other source will be reformatted later.


docs/Makefile
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Doxygen requires these packages:
+# doxygen (for doxygen)
+# texlive-full (for pdflatex and float.sty, used by PDF generation)
+
+TARGET=refman
+PDF_TARGET=$(TARGET)/latex/refman.pdf
```
manojgop: Do we need SRC (input directories for docxygen to generate documentation). Where should we add this.

danintel: Add source directories to the `INPUT` line in file `docs/Doxyfile`.
File Doxyfile is not shown because it is large--click on it to view the comments.
I'll add a comment to the Makefile.

Issue #310, opened 2020-03-09
Put hardcoded worker-lookup params in config


rranjan3:
Remove hardcoded fields like - application-type, org-id, etc. in config files. These should not be there in the code.

rranjan3:
Fixed. Configs created in tcf_connector.toml -
[WorkerConfig]
ApplicationTypeId
OrganizationId

PR #338, opened 2020-03-19
Move hardcoded worker attributes to config

rranjan3:
Address issue https://github.com/hyperledger/avalon/issues/310

### DIFF
--- a/sdk/avalon_sdk/tcf_connector.toml
+++ b/sdk/avalon_sdk/tcf_connector.toml
@@ -79,3 +79,9 @@ gas_price = "100"
 [fabric]
 fabric_network_file = "sdk/avalon_sdk/fabric/network.json"
 channel_name = "mychannel"
+
+[WorkerConfig]
+# These configs need to be same as that in tcs_config -
+# https://github.com/hyperledger/avalon/blob/master/config/tcs_config.toml
+ApplicationTypeId = "11aa22bb33cc44dd"
+OrganizationId = "aabbcc1234ddeeff"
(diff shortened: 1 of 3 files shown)


sdk/avalon_sdk/tcf_connector.toml
```diff
@@ -79,3 +79,7 @@ gas_price = "100"
 [fabric]
 fabric_network_file = "sdk/avalon_sdk/fabric/network.json"
 channel_name = "mychannel"
+
+[WorkerConfig]
```
manojgop: I assume these values should match with that present in https://github.com/hyperledger/avalon/blob/master/config/tcs_config.toml. Please add a comment in this file.

rranjan3: Done.

sdk/avalon_sdk/tcf_connector.toml
```diff
@@ -79,3 +79,9 @@ gas_price = "100"
...
+# These configs need to be same as that in tcs_config -
```
danintel: If they have to be the same, can't the program read them directly from c`onfig/tcs_config.toml` ?

rranjan3: The connector config has different set of configs specific to connector. So, it reads from a separate config file. It would not be a good approach to pass in 2 config files. Though these configurations might themselves be moved shortly because the connector will not handle worker sync-up. There would be a separate component doing this which would need the config. For now this is an added responsibility of the connector and hence the configs.

PR #342, opened 2020-03-20
Reinstate lint with module selection

rranjan3:
- Make lint script capable of doing selective lint only on modules passed
- Add Dockerfile to .gitattributes
- Update Blockchain connector Dockerfile to use make for build
- Fix lint issues in python modules


bin/run_lint
```diff
@@ -237,13 +237,16 @@ do
         ;;
     esac
 done
+if [ -z "$@" ]; then
+    echo "$0: No target MODULE(s) specified to lint"
```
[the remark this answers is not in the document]
rranjan3: Good idea @danintel . Done

rranjan3:
> Adding selective lint is good.
> But `run_lint` should default to checking all modules configured here (instead of erroring out).
> 
> I see someone has also disabled running `run_lint during builds. As a result, several lint errors have been introduced.
> 
> `run_lint` needs to be reenabled to run automatically during builds otherwise new lint errors will be continually introduced.

Yes I am making it part of the default build target. Not having it as a separate utility(though running independently is also perfectly fine). So, any build of code would be preceeded by a lint check which is the most apt time in the build process for static analysis.

rranjan3:
Addressed all your comments @danintel. Thanks

bin/run_lint
```diff
@@ -195,16 +195,28 @@ lint_module() {
...
+enclave_manager \
+common_python \
+common_crypto_utils \
+tc \
+common \
+docs_workload_tutorial \
+sdk \
+listener \
+shared_kv_storage \
+blockchain_connector
```
manojgop: Should we add examples/apps/generic_client either now or later ?

bin/run_lint
```diff
@@ -260,22 +276,53 @@ top_dir=$(cd $(dirname $(dirname $0)) && pwd)
 rv=0
 
 # Run pycodestyle or pylint or both for each module.
-# Modules should be ranked in decreasing likelihood of having lint errors.
 
 # Language:                                Python  | C
 # Checks:                           pycodestyle    | pycodestyle
 #                                           pylint |    cppcheck
 export PYTHONPATH=$PYTHONPATH:$top_dir/examples/common/python
```
manojgop: $top_dir/examples/common/python is not valid any more. Should be updated

blockchain_connector/Dockerfile
```diff
@@ -112,6 +116,8 @@ COPY ./common/python /project/avalon/common/python
 COPY ./common/crypto_utils /project/avalon/common/crypto_utils
 COPY ./sdk /project/avalon/sdk
 COPY ./blockchain_connector /project/avalon/blockchain_connector
+COPY ./.git /project/avalon/.git
+COPY ./.gitignore /project/avalon/
```
manojgop: Why are we copying .git and .gitignore in container image ? Reason is not obvious. Please add a comment

blockchain_connector/Dockerfile
```diff
@@ -124,25 +130,18 @@ RUN mkdir -p build \
...
 
 #install avalon crypto module
 WORKDIR /project/avalon/common/crypto_utils
 
 RUN echo "Building Avalon Common Crypto Python\n" \
-  && make \
-  && echo "Installing Avalon Common Crypto Python" \
-  && pip3 install dist/*.whl
-
+  && make
```
manojgop: Are we using crypto functionalities in block chain connector ?

listener/Dockerfile
```diff
@@ -1,161 +1,166 @@
...
+RUN apt-get update \
+ && apt-get install -y -q \
+    pkg-config \
+    cmake \
+    make \
+    libprotobuf-dev \
+    python3-dev \
+    python3-pip \
+    swig \
+    software-properties-common \
```
manojgop: git, pylint3 and pycodestyle are added in almost all Dockerfiles. Please create a staged build for lint images and reuse it.

bin/run_lint
rranjan3: Not needed as all installations using pip3 goes to default dist-packages

blockchain_connector/Dockerfile
rranjan3: Updated

blockchain_connector/Dockerfile
rranjan3: Not looking into it as of now for this lint specific PR.

listener/Dockerfile
rranjan3: Can I take up the optimization separately maybe once the CPP lint is also enabled?

blockchain_connector/Dockerfile
manojgop: Copying .git folder and .gitignore in container image looks bit odd. It's not good to expose specific git configurations and data in each container image. Why not create a separate dockerfile and docker-compose for lint which looks better than adding (read polluting ;-) ) each docker image with lint tools . For example refer -> https://github.com/hyperledger/sawtooth-core/blob/master/docker/compose/run-lint.yaml

rranjan3:
@manojgop replying to your opinion for separate docker for lint. 

In my opinion, lint should be an intergral part of code built. I would rather not build a code which has static analysis issues. Having separate dockerfile would defeat this. How can we make somebody run lint if he/she does a selective build of an image? To address this we would again have some workaround. 

Then also thinking from the modularilty point of view. Even if we decide to move modules to separate repos, this approach would not break the linting. Though some rework would still be needed to reuse the lint script we have.

Performance wise, lint is not a lot of overhead on build time though on missing out linting before a complete build or waiting on CI would be some significant turnaround time.

These are the reasons which made be choose this approach over any other. Still, if you feel putting the lints together for the reasons you stated, it can definitely be done.

rranjan3:
Updated @manojgop  to run as a separate docker-compose.

listener/Dockerfile
```diff
@@ -1,161 +1,161 @@
...
-RUN apt-get update \
- && apt-get install -y -q python3-pip \
- && pip3 install --upgrade json-rpc \
- && echo "Install Common Python, Crypto, Listener and SDK packages\n" \
- && pip3 install dist/*.whl \
- && echo "Remove unused packages from image\n" \
- && apt-get autoremove --purge -y -q python3-pip \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
+# Copyright 2020 Intel Corporation
```
manojgop: Why entire file is showing in the diff ? Some formatting got modified ?

shared_kv_storage/Dockerfile
```diff
@@ -1,98 +1,98 @@
...
-RUN apt-get update \
- && apt-get install -y -q python3-pip \
- && echo "Install Shared KV package\n" \
- && pip3 install dist/*.whl \
- && echo "Remove unused packages from image\n" \
- && apt-get autoremove --purge -y -q python3-pip \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
-
+# Copyright 2019 Intel Corporation
```
manojgop: Why entire file is showing in the diff ? Some formatting got modified ?

listener/Dockerfile
manju956: I guess, it could be due to the existing carriage return character (^M) in the listener Dockerfile.

rranjan3: Yes @manojgop , updated .gitattributes to include Dockerfile for eol native formatting.

rranjan3:
Example apps to be added for linting. Issue #350

manojgop:
> @manojgop replying to your opinion for separate docker for lint.
> 
> In my opinion, lint should be an intergral part of code built. I would rather not build a code which has static analysis issues. Having separate dockerfile would defeat this. How can we make somebody run lint if he/she does a selective build of an image? To address this we would again have some workaround.
> 
> Then also thinking from the modularity point of view. Even if we decide to move modules to separate repos, this approach would not break the linting. Though some rework would still be needed to reuse the lint script we have.
> 
> Performance wise, lint is not a lot of overhead on build time though on missing out linting before a complete build or waiting on CI would be some significant turnaround time.
> 
> These are the reasons which made be choose this approach over any other. Still, if you feel putting the lints together for the reasons you stated, it can definitely be done.

@rranjan3 Thanks for taking care of this. Approved PR

Jenkinsfile
```diff
@@ -61,6 +61,12 @@ pipeline {
             }
         }
 
+        stage('Run Lint') {
+            steps {
+                sh 'docker-compose -f docker-compose-lint.yaml up'
```
pankajgoyal2: @rranjan3 I assume that you have tested the negative scenario which CI system captures and fails.

manojgop: @pankajgoyal2 Good point. @rranjan3 Could you confirm that CI build will fail in case of lint error

rranjan3: The service(within docker-compose) does return an exit code 1 in case of failures. Should that suffice or need to add extra handling?

pankajgoyal2:
> @manojgop replying to your opinion for separate docker for lint.
> 
> In my opinion, lint should be an intergral part of code built. I would rather not build a code which has static analysis issues. Having separate dockerfile would defeat this. How can we make somebody run lint if he/she does a selective build of an image? To address this we would again have some workaround.
> 
> Then also thinking from the modularity point of view. Even if we decide to move modules to separate repos, this approach would not break the linting. Though some rework would still be needed to reuse the lint script we have.
> 
> Performance wise, lint is not a lot of overhead on build time though on missing out linting before a complete build or waiting on CI would be some significant turnaround time.
> 
> These are the reasons which made be choose this approach over any other. Still, if you feel putting the lints together for the reasons you stated, it can definitely be done.

As you are making running lint through seperate docker-compose file as part of CI system, CI system only build SGX SIM mode. So, developers need to run this for each and every commit that they make to ensure lint issues are captured for every scenario which CI might not catch.

Jenkinsfile
pankajgoyal2: I suggest, please introduce some lint specific vulnerability intentionally and run it...it should exit without even going to Build stage.

rranjan3: Updated @pankajgoyal2 to handle that. There's a flag specifically to pass out the return status outside docker-compose - https://docs.docker.com/compose/reference/up/

rranjan3:
> > @manojgop replying to your opinion for separate docker for lint.
> > In my opinion, lint should be an intergral part of code built. I would rather not build a code which has static analysis issues. Having separate dockerfile would defeat this. How can we make somebody run lint if he/she does a selective build of an image? To address this we would again have some workaround.
> > Then also thinking from the modularity point of view. Even if we decide to move modules to separate repos, this approach would not break the linting. Though some rework would still be needed to reuse the lint script we have.
> > Performance wise, lint is not a lot of overhead on build time though on missing out linting before a complete build or waiting on CI would be some significant turnaround time.
> > These are the reasons which made be choose this approach over any other. Still, if you feel putting the lints together for the reasons you stated, it can definitely be done.
> 
> @rranjan3 Thanks for taking care of this. Approved PR



> Example apps to be added for linting. Issue #350

Updated the example directory here itself @manojgop .

bin/run_lint
```diff
@@ -259,23 +271,60 @@ fi
...
+                lint_module listener                 "$SINCE" 1  0   0  0 || rv=1
+                ;;
+        shared_kv_storage)
+                lint_module shared_kv_storage        "$SINCE" 1  0   0  1 || rv=1
+                ;;
+        blockchain_connector)
+                lint_module blockchain_connector     "$SINCE" 1  0   0  0 || rv=1
+                ;;
+        examples)
+                lint_module examples                 "$SINCE" 1  0   0  1 || rv=1
```
manojgop: Is entire code under examples directory lint compliant ? There are lot of applications including 3rd party applications in examples directory

rranjan3: The lint script would do a nested search for all .py and .cpp files and run lint on them.

docker-compose-lint.yaml
```diff
@@ -0,0 +1,37 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '3.5'
+
+services:
+  avalon-lint:
+    container_name: avalon-lint
```
swastik-goyal: Remove container_name from compose file if you want to run this as part of CI

pankajgoyal2: Remove container_name if you want to run it as part of CI

rranjan3: Thanks @pankajgoyal2, nice catch. Fixed.
Tested with error case as well @manojgop. Does fail the Jenkins build.

PR #345, opened 2020-03-23
Handle multiple requester clients

rranjan3:
- Bugfix to handle multiple work order requester for Ethereum
This commit handles multiple work order requests from multiple requesters but not simultaneously. The 'nonce' field in any transaction needs to be unique and equal to one more than the number of transactions submitted with the account being used. This needs a synchronization between the requesters by some means. In a future implementation, some central counter generator service could be used which would also hold multiple signing accounts and return counter for account requested.
But with the current implementation only a single account is being considered.

### DIFF
--- a/sdk/avalon_sdk/ethereum/ethereum_listener.py
+++ b/sdk/avalon_sdk/ethereum/ethereum_listener.py
@@ -76,13 +76,23 @@ class EventProcessor:
             callback(event, *kargs, **kwargs)
             self.queue.task_done()
 
-    async def sync_handler(self, *kargs, **kwargs):
+    async def sync_handler(self, check_event_callback=None, *kargs, **kwargs):
         """Start a synchronous event handler to handle an event."""
         logging.info("Started synchronous handler to handle an event")
-        event = await self.queue.get()
-        logging.debug("Event popped from listener queue")
+        while True:
 
-        self.queue.task_done()
+            event = await self.queue.get()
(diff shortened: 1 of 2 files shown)


sdk/avalon_sdk/ethereum/ethereum_listener.py
```diff
@@ -66,12 +66,20 @@ def __init__(self, config):
...
+
+            event = await self.queue.get()
+            logging.debug("Event popped from listener queue")
+            self.queue.task_done()
+
+            # Break out of the loop if check_event_callback method is
+            # defined and returns True when called.
+            if check_event_callback is not None and \
+                    check_event_callback(event, *kargs, **kwargs) is True:
+                break
```
manojgop: If check_event_callback  is None is this an error  case? If yes, return an error. If not, please explain why.

sdk/avalon_sdk/ethereum/ethereum_work_order.py
```diff
@@ -220,6 +221,28 @@ def encryption_key_get(self, worker_id, requester_id,
         pass
 
 
+def is_wo_id_in_event(event, *kargs, **kwargs):
```
Ram-srini: Change function to private

rranjan3: It is a static method @Ram-srini , since we are not really dealing with member variables here.

PR #346, opened 2020-03-23
Add Avalon Python SDK to Doxygen

danintel:

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -25,12 +25,18 @@ VERSION=$$(../bin/get_version)
 # List source directories and files to be documented by Doxygen here.
 # Doxyfile's RECURSIVE=YES makes it also use subdirectories.
 INPUT=	refman/README.md \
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -25,12 +25,18 @@ VERSION=$$(../bin/get_version)
 # List source directories and files to be documented by Doxygen here.
 # Doxyfile's RECURSIVE=YES makes it also use subdirectories.
 INPUT=	refman/README.md \
-	../common/sgx_workload
+	../common/sgx_workload \
+	../sdk
+EXCLUDE= test \
```
manju956: Do we have module named test?

danintel: No directory named `test` under `sdk`, but we have one under `examplles/common/java` and I want to leave it in case I decide to add it to Doxygen.

manju956: Ok

PR #348, opened 2020-03-24
Add validation in listener to fix following bugs

Ram-srini:
1. TCFH-733 - Avalon is processing the null data and data hash input request in the indata section and giving incorrect results
2. TCFH-734 - WorkOrderGetResult fails with server error with the removal of WorkOrderid
3. TCFH-736 - responseTimeoutMSecs parameter in WorkOrderSubmit API is not working with special/ alphanumeric characters
4. TCFH-735 - WorkOrderSubmit fails with server error with the removal of requesterId
5. TCFH-703 - Direct model workOrderSubmit request is working without payload format in params field

### DIFF
--- a/listener/avalon_listener/tcs_listener.py
+++ b/listener/avalon_listener/tcs_listener.py
@@ -114,12 +114,7 @@ class TCSListener(resource.Resource):
             input_json = json.loads(input_json_str)
         except Exception as err:
             logger.exception("exception loading Json: %s", str(err))
-            response = {
-                "error": {
-                    "code": WorkOrderStatus.INVALID_PARAMETER_FORMAT_OR_VALUE,
-                    "message": "Error: Improper Json. Unable to load",
-                },
-            }
+            response["error"]["message"] = "Improper Json request"
             return response
 
         logger.info("Received request: %s", input_json['method'])
--- a/listener/avalon_listener/tcs_work_order_handler.py
+++ b/listener/avalon_listener/tcs_work_order_handler.py
@@ -15,6 +15,7 @@
 import time
 import json
 import logging
+
 from error_code.error_status import WorkOrderStatus
 from error_code.enclave_error import EnclaveError
 from avalon_sdk.direct.jrpc.jrpc_util import JsonRpcErrorCode
@@ -55,7 +56,6 @@ class TCSWorkOrderHandler:
         self.workorder_count = 0
         self.max_workorder_count = max_wo_count
         self.workorder_list = []
-
         self.__work_order_handler_on_boot()
 
 # ---------------------------------------------------------------------------------------------
@@ -88,6 +88,18 @@ class TCSWorkOrderHandler:
                 self.workorder_list.append(wo_id)
                 self.workorder_count += 1
 
+    def __is_worker_exists(self, worker_id):
+        """
+        Function to check if worker is exists or not
+        Returns
+            True if exists or False if not
+        """
+        logger.info("worker id to check in lmdb {}".format(worker_id))
+        if self.kv_helper.get("workers", worker_id):
+            return True
+        else:
+            return False
+
 # ---------------------------------------------------------------------------------------------
     def WorkOrderGetResult(self, **params):
         """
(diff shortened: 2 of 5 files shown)


manojgop:
Can you please mention what exactly is the issue fixed in commit message than only mentioning BUG IDs. these are internal jira and doesn't give enough context to a reviewer

listener/avalon_listener/tcs_listener.py
```diff
@@ -104,6 +104,29 @@ def __init__(self, config):
...
+        for k in input_json.keys():
+            if k not in valid_keys:
+                logger.info("Invalid parameter in json rpc request")
+                return False
+        if "jsonrpc" in input_json:
+            if input_json["jsonrpc"] != "2.0":
+                logger.info("Invalid json rpc version")
+                return False
+        if "id" in input_json:
+            return type(input_json["id"]) == int
```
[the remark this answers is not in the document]
Ram-srini: You mean to log it?

listener/avalon_listener/tcs_listener.py
```diff
@@ -104,6 +104,29 @@ def __init__(self, config):
         for m in rpc_methods:
             self.dispatcher.add_method(m)
 
+    def _valid_json_rpc_request(self, input_json):
```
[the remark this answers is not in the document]
Ram-srini: We are using jsonrpc package. It has JSONRPCResponseManager, but it is not throwing any exception for wrong json rpc methods.

listener/avalon_listener/tcs_listener.py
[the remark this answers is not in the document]
pankajgoyal2: yes

sdk/avalon_sdk/worker/worker_details.py
```diff
@@ -205,9 +207,10 @@ def load_worker(self, worker_data):
             self.proof_data = json.loads(
                 worker_data['workerTypeData']['proofData'])
 
-        self.worker_id = crypto_utility.strip_begin_end_public_key(
+        w_id = crypto_utility.strip_begin_end_public_key(
             worker_data['workerTypeData']['verificationKey']) \
-            .encode("UTF-8").hex()
+            .encode("UTF-8")
+        self.worker_id = hashlib.sha256(w_id).hexdigest()
```
manojgop: Could you please add a comment what is this change for ? @rranjan3 had a similar done a change in a different PR.

listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -99,42 +113,52 @@ def WorkOrderGetResult(self, **params):
...
+            if value:
+                response = json.loads(value)
+                if 'result' in response:
+                    return response['result']
+
+                # response without a result should have an error
+                err_code = response["error"]["code"]
+                err_msg = response["error"]["message"]
+                if err_msg == self.INVALID_WORKLOAD_MESSAGE:
+                    err_msg = 'Invalid workload id'
```
manojgop: Better to check the err_code and then set the required err_msg.
```
 if err_code == xxx:
     err_msg = 'Invalid workload id'
```

listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -55,7 +56,8 @@ def __init__(self, kv_helper, max_wo_count):
         self.workorder_count = 0
         self.max_workorder_count = max_wo_count
         self.workorder_list = []
-
+        self.INVALID_WORKLOAD_MESSAGE = \
+            'CreateWorkloadProcessor function returned null'
```
manojgop: This doesn't look good. listener ideally shouldn't check for exact error message string of enclave manager. Instead use an error code.

listener/avalon_listener/tcs_listener.py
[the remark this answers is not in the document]
manojgop: That's interesting.  In above link , the implementation pf `JSONRPCResponseManager.handle()`  checks for `JSONRPCInvalidRequestException` and `return JSONRPC20Response(error=JSONRPCInvalidRequest()._data)`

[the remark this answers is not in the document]
manojgop: Looks like we are not checking for error in response object in function `_process_request()`. Check if response object has error attribute or only data attribute
```
response = JSONRPCResponseManager.handle(data, self.dispatcher)
return response.data
```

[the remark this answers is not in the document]
Ram-srini: These exceptions should be thrown from handler functions, not outside. Otherwise it will return in http error 500.

listener/avalon_listener/tcs_work_order_handler.py
Ram-srini: This is what we get it from enclave manager. value error and message as CreateWorkloadProcessor function returned null

listener/avalon_listener/tcs_work_order_handler.py
Ram-srini: We won't get the error code from enclave manager for this error.

listener/avalon_listener/tcs_listener.py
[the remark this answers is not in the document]
manojgop: Yes.  Need to check if response object returned by `JSONRPCResponseManager.handle()` has error attribute if JSON RPC request is invalid and then we should handle the scenario for error. It's better to use library provided error handling for malformed JSON RPC request  than writing a custom function

listener/avalon_listener/tcs_work_order_handler.py
manojgop: Listener need to check only for error code enclave manager. It should not compare the error message from enclave manager. If you need specific error code then we need to modify enclave manager to return the specific error code

listener/avalon_listener/tcs_work_order_handler.py
manojgop: Thought so. @manju956 Can we return a new error code from enclave manager for this scenario. Please suggest where to where to make this change.

manju956: Enclave manager don't add any error code as such. New error codes can be added in tcf_error.h and we need to define a class for the new error code in error.h.
Also, we need to update the enclave code to return new error code when invalid workload_id has been identified by the enclave.

manojgop: Yeah. We need to do define new error codes corresponding to workload id and enclave manager should return that. This should be checked by listener

listener/avalon_listener/tcs_listener.py
manojgop: Please check https://www.jsonrpc.org/specification. An identifier established by the Client that MUST contain a String, Number, or NULL value if included. 

There are other error conditions for json rpc request which are not validated by this function. I would suggest to use library provided error handling than writing a custom error handling function for validating json rpc request. 
`response = JSONRPCResponseManager.handle()` returns a response object which will contain an error field in case of error. For invalid json rpc request,  `response.error` will look like this -> `{'code': -32600, 'message': 'Invalid Request'}`

Ram-srini: In our spec everywhere id is mentioned as integer. Then we should update our spec as well.

sdk/avalon_sdk/worker/worker_details.py
Ram-srini: Done

listener/avalon_listener/tcs_listener.py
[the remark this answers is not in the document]
Ram-srini: not required

sdk/avalon_sdk/worker/worker_details.py
manojgop: The worker_id stored in block chain as part of worker details should already be in 32 bytes as enclave manager is storing it in 32 bytes in shared KV DB -> https://github.com/hyperledger/avalon/blob/master/enclave_manager/avalon_enclave_manager/enclave_manager.py#L99. So why we need to duplicate the calculation of worker id again ?

manojgop: load_worker() function should take worker_id as argument instead of recalculating it here

Ram-srini: why do we need to pass worker_id to load_worker again? worker_data['workerTypeData']['verificationKey'] which is enclave public key, that we are converting it to sha256. Don't you think it is also duplicate? First of all why worker_id calculation is there in load_worker() earlier as well? Earlier also it was duplication calculation.

Ram-srini: Neater approach would be to use the worker_id from the worker_lookup() API rather than from the worker_details?

manojgop: Verification Key is the public part of signing key which is different from worker id.  So verification key is not duplicate. Earlier logic was also not correct. Can we fix that along with this bug fix PR ? Else please add a TODO comment in code and raise an issue which can be fixed in different PR.

Ram-srini: in load_worker(), we are calculating worker_id from ['workerTypeData']['verificationKey']. Why it was not taken care earlier itself? How come it is fine now if that logic is incorrect?

manojgop: Pass worker_id in constructor of workerDetails

Ram-srini: Created the issue to remote the worker id calculation in the sdk. https://github.com/hyperledger/avalon/issues/359

Ram-srini: > Pass worker_id in constructor of workerDetails

No need to worker_id in workerDetails. User should use the worker id got from worker_lookup().

manojgop: I mean when we crate WorkerDetails object, we should know for which worker_id this object is being created for. For that we can use worker_id we got from worker lookup.

Issue #292, opened 2020-02-25
Avalon should not require both solc and solcx


danintel:
Recently Avalon was modified to require the solcx compiler.
However, it also requires the solc compiler (otherwise applications will not run without both).

Modify Avalon to require just solcx and not both solc and solcx.

rranjan3:
Fixed with #349.

rranjan3:
@danintel I think this issue can be closed now.

PR #349, opened 2020-03-24
Cleanup redundant solc installation

rranjan3:
- Remove solc apt-get installation
- Rely only py-solc-x which suffices for Solidity handling
  in Python
- Update documentations

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -153,7 +153,7 @@ The steps below will set up a Python virtual environment to run Avalon.
 10. Install PIP3 packages into your Python virtual environment:
 
     ```bash
-    pip3 install --upgrade setuptools json-rpc py-solc py-solc-x web3 colorlog twisted wheel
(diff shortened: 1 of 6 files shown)


danintel:
Fixes Issue #292

rranjan3:
> 1. Remove these lines in sdk/avalon_sdk/tcf_connector.toml as they are not used:
> 
> ```
> # Version of solc to be used for compiling Solidity contracts
> solc_version = "v0.5.15"
> ```
> 
> 1. Change solc to solcx in docker/Dockerfile:
> 
> ```
> # Install specific version of solc to support versions <0.6.0
> ```
> 
> 1. Change solc to solcx in blockchain_connector/Dockerfile:
> 
> ```
> Install specific version of solc to support versions <0.6.0
> ```

Thanks @danintel for the suggestions.
But `solc` is right in the above contexts. It is always the solc which would do the compilation and we maintain specific version of it. `solcx` is just a wrapper over it. The one being removed is a debian package installed using apt-get.

PR #365, opened 2020-03-27
Handle multiple requester clients

rranjan3:
- Bugfix to handle multiple work order requester for Ethereum
This commit handles multiple work order requests from multiple requesters
but not simultaneously. The 'nonce' field in any transaction needs to be
unique and equal to one more than the number of transactions submitted
with the account being used. This needs a synchronization between the
requesters by some means. In a future implementation, some central counter
generator service could be used which would also hold multiple signing
accounts and return counter for account requested. Refer https://github.com/ethereum/wiki/wiki/Glossary to
look for Account nonce.
But with the current implementation only a single account is being considered.


rranjan3:
Raise another PR with separate commit.

PR #367, opened 2020-03-27
Add data field in work order status response

Ram-srini:
As per the EEA spec section 6, work order status payload should have work order id in error tag.

### DIFF
--- a/listener/avalon_listener/tcs_work_order_handler.py
+++ b/listener/avalon_listener/tcs_work_order_handler.py
@@ -113,11 +113,17 @@ class TCSWorkOrderHandler:
         """
         if "workOrderId" in params:
             wo_id = params["workOrderId"]
+            # Work order status payload should have
+            # data field with work order id as defined in EEA spec section 6.
+            data = {
+                "workOrderId": wo_id
+            }
             if not is_valid_hex_str(wo_id):
                 logging.error("Invalid work order Id")
                 raise JSONRPCDispatchException(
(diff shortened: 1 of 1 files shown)


listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -113,11 +113,17 @@ def WorkOrderGetResult(self, **params):
         """
         if "workOrderId" in params:
             wo_id = params["workOrderId"]
+            # Work order status payload should have
+            # data filed with work order id as defined in EEA spec section 6.
```
manojgop: filed -> field

listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -172,14 +184,19 @@ def WorkOrderSubmit(self, **params):
...
+        }
         req_validator = WorkOrderRequestValidator()
         valid, err_msg = req_validator.validate_parameters(
             input_value_json["params"])
         if not valid:
             raise JSONRPCDispatchException(
                 JsonRpcErrorCode.INVALID_PARAMETER,
-                err_msg
+                err_msg,
+                data=data
```
manojgop: Any specific reason why you used data=data instead of just passing data.

Ram-srini: JSONRPCDispatchException has default arguments and variable arguments, it is named argument to that function.

PR #370, opened 2020-03-27
Add workOrderId field for work order Error response from enclave

manju956:

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
@@ -446,6 +446,15 @@ namespace tcf {
         JsonSetNumber(error, "code", err_code, "failed to serialize error code");
         JsonSetStr(error, "message", err_message, "failed to serialize error message");
 
+        jret = json_object_set_value(error, "data", json_value_init_object());
+        tcf::error::ThrowIf<tcf::error::RuntimeError>(
+            jret != JSONSuccess, "failed to serialize data");
+        JSON_Object* data = json_object_get_object(error, "data");
+        tcf::error::ThrowIfNull(data, "failed to serialize the data object");
+
+        JsonSetStr(data, "workOrderId", work_order_id.c_str(),
+            "failed to serialize work order id");
+
         // Serialize the resulting json
         size_t serializedSize = json_serialization_size(resp_value);
         ByteArray serialized_response;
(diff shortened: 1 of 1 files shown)


listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -131,14 +131,18 @@ def WorkOrderGetResult(self, **params):
                 err_code = response["error"]["code"]
                 err_msg = response["error"]["message"]
 
+                data = {
```
Ram-srini: This is taken care in PR#367

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -435,6 +435,9 @@ namespace tcf {
         JsonSetStr(resp, "jsonrpc", "2.0", "failed to serialize jsonrpc");
         JsonSetNumber(resp, "id", json_request_id, "failed to serialize json request id");
 
+        JsonSetStr(resp, "workOrderId", work_order_id.c_str(),
```
Ram-srini: "workOrderId" should be inside "data". It should be like
{
"jsonrpc": "2.0",
"id": <jrpc_request_id>,
"error": {
"code": <error_code>,
"message": <error_message>,
"data": {
"workOrderId": <work_order_id
}
}

manju956: Done

listener/avalon_listener/tcs_work_order_handler.py
manju956: I shall remove this once PR#367 is merged.

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -444,6 +444,8 @@ namespace tcf {
 
         JsonSetNumber(error, "code", err_code, "failed to serialize error code");
         JsonSetStr(error, "message", err_message, "failed to serialize error message");
+        JsonSetStr(error, "workOrderId", work_order_id.c_str(),
```
Ram-srini: workOrderId should be inside data field as below
"data": {
"workOrderId": <work_order_d>
}

manju956: Ok, will make the correction

PR #375, opened 2020-03-31
Generate EncryptionKeySignature and restructure enclave signup process

manju956:
Generate EncryptionKeySignature and restructure enclave signup process

- EncryptionKeySignature is generated as part of enclave data
- EncryptionKeySignature is made as part of public enclave data
- EnclaveData class is made as singleton so that worker enclave stores the context of enclave data. 
   This avoids sealing of enclave data.
- Multiple APIs which expects sealed data are updated to remove the parameter

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
@@ -57,6 +57,9 @@ EnclaveData::EnclaveData(void) {
     // Create the public encryption key
     public_encryption_key_ = private_encryption_key_.GetPublicKey();
 
+    // Create encryption key signature
+    generate_encryption_key_signature();
+
     SerializePrivateData();
     SerializePublicData();
 }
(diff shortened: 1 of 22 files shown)


tc/sgx/trusted_worker_manager/enclave/enclave_data.h
```diff
@@ -65,7 +69,11 @@ class EnclaveData {
     // allocated storage for the sealed data. There are a number of ways
     // to do this; for now, the constant will be good enough.
     static const size_t cMaxSealedDataSize = 8192;
-    static const size_t cMaxPublicDataSize = 4096;
+    /*
+       size of the base64 encoded encryption key signature is 2733
+       formula to calculate size - (no of bytes * 4 / 3) + 2 + 1
+    */
+    static const size_t cMaxPublicDataSize = 6829;
```
manojgop: calculate the size programatically based on above comment instead of hard coding the value. Also calculate the base64 encoded encryption key signature.  Update the comment to be more elaborate like why are we adding 2+1 in calculation above (mention about double "=" for padding and extra byte to store null character).

manju956: the encryption_key_signature variable stores the base64 encoded version. I will add more description to the formula.

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -57,6 +57,9 @@ EnclaveData::EnclaveData(void) {
     // Create the public encryption key
     public_encryption_key_ = private_encryption_key_.GetPublicKey();
 
+    /* Create encryption key signature */
```
danintel: For consistency, use the same comment style as the rest of the file.  That is:
`//`

manju956: Ok

tc/sgx/trusted_worker_manager/enclave/enclave_data.cpp
```diff
@@ -220,6 +136,7 @@ void EnclaveData::SerializePrivateData(void) {
     b64_public_encryption_key.clear();
 
     size_t serializedSize = json_serialization_size(dataValue);
+    //cMaxSealedDataSize = serializedSize;
```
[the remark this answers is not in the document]
manju956: Done

PR #376, opened 2020-03-31
Update cleanup script

manju956:
Source paths of many modules in the cleanup script are updated to
the latest. This script will be useful in cleaning up build artifacts
in non-docker builds.

### DIFF
--- a/tools/clean.sh
+++ b/tools/clean.sh
@@ -27,31 +27,48 @@ SRCDIR="$(realpath ${SCRIPTDIR}/..)"
 rm -f $SRCDIR/enclave.pem
 
 # Remove generated ias-certificate file
-rm -f $SRCDIR/tc/sgx/common/crypto/verify_ias_report/ias-certificates.cpp
+rm -f $SRCDIR/common/cpp/verify_ias_report/ias-certificates.cpp
 
 # Remove LMDB files
 echo "******** DELETE LMDB FILES **************"
 rm -f $SRCDIR/config/Kv_Shared*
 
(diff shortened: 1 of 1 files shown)


tools/clean.sh
```diff
@@ -27,31 +27,48 @@ SRCDIR="$(realpath ${SCRIPTDIR}/..)"
...
-cd $SRCDIR/common/python
-make clean
+# --------------- COMMON CPP ---------------
+cd $SRCDIR/common/cpp
+rm -rf build
 
-# --------------- ENCLAVE MANAGER ---------------
-cd $SRCDIR/enclave_manager
-SGX_MODE=${SGX_MODE:-SIM} make clean
+# --------------- SGX WORKLOAD ---------------
```
danintel: change SGX to INTEL SGX

tools/clean.sh
```diff
@@ -27,31 +27,48 @@ SRCDIR="$(realpath ${SCRIPTDIR}/..)"
...
+cd $SRCDIR/common/sgx_workload
+rm -rf build
 
-# --------------- COMMON SGX WORKLOAD ---------------
-rm -rf $SRCDIR/common/sgx_workload/build
+# --------------- SGX COMMON ---------------
```
danintel: change SGX COMMON to INTEL SGX COMMON

tools/clean.sh
manju956: Done

tools/clean.sh
manju956: Done

PR #377, opened 2020-03-31
Add validation to work order submit request parameters

Ram-srini:
Fix to address the following jira bugs
1. TCFH-745 - "Server Error" message appears on passing invalid indata without the index present in json.
2. TCFH-746 - Data is getting processed when “workerencryptionkey” parameter has special characters.


common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -269,21 +270,17 @@ def generate_client_signature(
...
-            nonce = crypto.string_to_byte_array(
-                input_json_params["requesterNonce"])
-        else:
+        if "requesterNonce" in input_json_params and \
+                len(input_json_params["requesterNonce"]) == 0:
             # [NO_OF_BYTES] 16 BYTES for nonce.
             # This is the recommendation by NIST to
             # avoid collisions by the "Birthday Paradox".
-            nonce = crypto.random_bit_string(NO_OF_BYTES)
+            input_json_params["requesterNonce"] = secrets.token_hex(16)
```
danintel: Change 16 to NO_OF_BYTES

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -269,21 +270,17 @@ def generate_client_signature(
...
 
-        request_nonce_hash = crypto.compute_message_hash(nonce)
-        nonce_hash = (crypto.byte_array_to_base64(request_nonce_hash)).encode(
-            'UTF-8')
         hash_string_1 = self.__calculate_hash_on_concatenated_string(
-            input_json_params, nonce_hash)
+            input_json_params, input_json_params["requesterNonce"].encode(
+                'UTF-8'
+            ))
```
manojgop, line 283: Shouldn't we use hash of nonce ? Trusted compute spec v1.1 mentions "Generate a random number and store its hash in requesterNonce" in section 6.6.1 Request Hash Calculation and Encryption

Ram-srini: requesterNonce is a random string generated by the participant. At line 278, creating random hex string.

Ram-srini: @manojgop Earlier steps are not correct. we need to send the random string in requesterNonce and hash will calculated on the string and used in request hash calculation.

manojgop: Need to modify the function `__calculate_hash_on_concatenated_string()` . Update the function param from nonce_hash to nonce. Also the function `__calculate_hash_on_concatenated_string()`  should be modified to calculate nonce hash from nonce as per spec 6.6.1. This function takes nonce (instead of noce_hash) as argument now. Calculate the nonce_hash within this function and concatenate it for final hash value calculation.

- Generate a random number and store its hash in requesterNonce.

- Calculate a hash value of the string concatenating the following values: requesterNonce, workOrderId, workerId, workloadId, and requesterId.

Ram-srini: Those steps are taken in out side of __calculate_hash_on_concatenated_string(), in generate_client_signature().

Ram-srini: requesterNonce is a random string generated by the participant. It is used to calculate a hash of this work order request. generate_client_signature() is shouldn't override that, if requesterNonce is passed in the request use that otherwise generate random string and use it in generate_client_signature(), it is taken care in that function.

Ram-srini: @manojgop I have review comments. Please review.

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -269,21 +269,18 @@ def generate_client_signature(
             input_json_params, session_key,
             session_iv, worker.encryption_key, data_key, data_iv)
 
-        if input_json_params["requesterNonce"] and \
-                is_valid_hex_str(input_json_params["requesterNonce"]):
-            nonce = crypto.string_to_byte_array(
-                input_json_params["requesterNonce"])
-        else:
+        if "requesterNonce" in input_json_params and \
+                len(input_json_params["requesterNonce"]) == 0:
```
manojgop, line 273: If user passes "requesterNonce" then shouldn't we add a check for valid hex string ?

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -269,21 +269,20 @@ def generate_client_signature(
...
+        if "requesterNonce" in input_json_params:
+            if len(input_json_params["requesterNonce"]) == 0:
+                # [NO_OF_BYTES] 16 BYTES for nonce.
+                # This is the recommendation by NIST to
+                # avoid collisions by the "Birthday Paradox".
+                input_json_params["requesterNonce"] = secrets.token_hex(
+                    NO_OF_BYTES)
+            elif not is_valid_hex_str(input_json_params["requesterNonce"]):
+                raise Exception("Invalid data format for requesterNonce")
+
```
manojgop, line 281: requesterNonce is a mandatory field and if it's not present in input request
LOG error and return fail status

Ram-srini: generate_client_signature() is utility function not an API. It is doing too many things, It is validating parameters, updating input request with many parameters like requesterNonce, encryptedRequestHash, encryptedSessionKey, verifyingKey. Ideally should update only requesterSignature as function name indicates.

manojgop: Yeah. That's correct. We should clean this by doing some parameter validation outside the function. Could be done later I suppose.

sdk/avalon_sdk/work_order/work_order_request_validator.py
```diff
@@ -104,6 +106,14 @@ def validate_parameters(self, params):
         if "encryptedRequestHash" in params and \
                 not is_valid_hex_str(params["encryptedRequestHash"]):
             return False, "Invalid data format for encrypted request hash"
+        if not is_valid_hex_str(params["requesterNonce"]):
```
manojgop, line 109: If "requesterNonce" field is not present will this return False ?

PR #380, opened 2020-04-01
Change SGX to Intel SGX in Comments and Messages

danintel:
Also break up long lines and make other minor corrections.


danintel:
Merged with #379.

PR #387, opened 2020-04-02
BUILD.md: Add `git clone` Instruction for Docker Builds

danintel:
A `git clone` is needed for Docker builds, not just
standalone builds.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -2,6 +2,7 @@
 Licensed under Creative Commons Attribution 4.0 International License
 https://creativecommons.org/licenses/by/4.0/
 -->
+
 # Building Hyperledger Avalon
 
 In order to build, install, and run Hyperledger Avalon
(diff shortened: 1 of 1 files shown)


BUILD.md
```diff
@@ -151,36 +153,36 @@ The steps below will set up a Python virtual environment to run Avalon.
...
-11. Build and install Avalon components:
+10. Build and install Avalon components:
 
     ```bash
     make clean
     make
     ```
 
-12. Build the Avalon SDK Python module:
+11. Build the Avalon SDK Python module:
```
pankajgoyal2, line 169: Steps 11 (building Avalon SDK python module), 12(building shared_kv), 13(Building Avalon listener) should be removed from here. Step 10 above takes care of that.

Another point (not a problem with your changes) is that in tools/rebuild.sh, we should use "make && make install"

danintel: That was added recently (adding sdk, shared_kv_storage, and listener) to `tools/rebuild.sh`
I will try that tomorrow.

danintel: Done

PR #390, opened 2020-04-03
Patch for v0.5 release

manojgop:
Below patches are applied for v0.5 release

- Update link to download OpenSSL source code
- Update Build documentation


manojgop:
> Update common/cpp/README.md alos. It also has reference to [www.openssl.org](http://www.openssl.org)

@Ram-srini That file only mentions openSSL 1.1. So I don't need to update that as part of this.

danintel:
I added PR #391 to document this patch release in BUILD.md

PR #405, opened 2020-04-08
Cleanup Ethereum generic client

rranjan3:
-Remove fabric related code from ethereum eneric client.
-Enable direct invocation from same client

### DIFF
--- a/examples/apps/generic_client/eth_generic_client.py
+++ b/examples/apps/generic_client/eth_generic_client.py
@@ -42,8 +42,6 @@ import avalon_crypto_utils.signature as signature
 from error_code.error_status import SignatureStatus
 from avalon_sdk.work_order_receipt.work_order_receipt \
     import WorkOrderReceiptRequest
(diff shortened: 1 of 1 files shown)


examples/apps/generic_client/eth_generic_client.py
```diff
@@ -78,8 +76,7 @@ def _parse_command_line(args):
         "-b", "--blockchain",
         help="Blockchain type to use in proxy model",
         choices={"fabric", "ethereum"},
```
manju956, line 78: Can fabric be removed from choices?

rranjan3: Deliberately left there @manju956 .  Pretty soon all the generic client would be merged. Then this would make sense. The other pieces of code removed in this PR were stale.

manju956: Thanks for the clarity.

PR #416, opened 2020-04-14
Basic KME listener

rranjan3:
- Add basic KMEListener that expects JRPC requests
- Minor bug fix in tcs_listener.py
- Add ErrorCodes

### DIFF
--- a/common/python/listener/base_jrpc_listener.py
+++ b/common/python/listener/base_jrpc_listener.py
@@ -1,0 +1,224 @@
+# Copyright 2020 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Values from -32768 to -32000 are reserved for pre-defined errors in
+# the JSON RPC spec.
+# 0 - success
+# 1 - unknown error
+# 2 - invalid parameter format or value
+# 3 - access denied
+# 4 - invalid signature
+# 5 - no more lookup results
+# 6 - unsupported mode (e.g. synchronous, asynchronous, pull, or notification)
+# 7 - busy
+
+
+import os
+import sys
+import json
+
+from urllib.parse import urlsplit
+from twisted.web import server, resource, http
+from twisted.internet import reactor, error as reactor_error
+from error_code.error_status import JRPCErrorCodes
+import utility.jrpc_utility as jrpc_utility
+
+
+from jsonrpc.dispatcher import Dispatcher
+from jsonrpc import JSONRPCResponseManager
+
+import logging
+logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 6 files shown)


enclave_manager/avalon_enclave_manager/kme_listener.py
```diff
@@ -0,0 +1,290 @@
...
+    return host_name, port
+
+# -----------------------------------------------------------------
+# -----------------------------------------------------------------
+
+
+def get_config_dir():
+    """
+    Returns the avalon configuration directory based on the
+    TCF_HOME environment variable (if set) or OS defaults.
```
danintel, line 248: Change the comment. to remove ` or OS defaults` (because reading the code there is no default value if `TCF_HOME` is not set).

enclave_manager/avalon_enclave_manager/kme_listener.py
```diff
@@ -0,0 +1,290 @@
...
+from error_code.error_status \
+    import WPERegisterStatus, WOPreprocessStatus, JRPCErrorCodes
+from utility.jrpc_utility import create_error_response
+
+logger = logging.getLogger(__name__)
+
+TCFHOME = os.environ.get("TCF_HOME", "../../")
+
+
+class KMEListener(resource.Resource):
```
manojgop, line 40: To avoid duplicate code, please create a BaseListener class in common/python package and let Avalon Listener and KME Listener inherit from BaseListener

enclave_manager/avalon_enclave_manager/kme_listener.py
```diff
@@ -0,0 +1,129 @@
...
+        self.config = config
+        self.dispatcher = Dispatcher()
+        self.kme_request_handler = None
+
+        # TODO : Uncomment following statements after
+        # a request handler is created
+        '''self.dispatcher.add_method(self.kme_request_handler\
+            .RegisterWorkorderProcessor)
+        self.dispatcher.add_method(self.kme_request_handler\
+            .PreporocessWorkorder)'''
```
manojgop, line 51: `config`, `dispatcher` could be used in base class. Also create `Dispatcher()` in base class and call `dispatcher.add_method()` in base class . Pass the array with rpc method names to base class constructor and add the methods in dispatcher in a loop.

```
class BaseJRPCListener(resource.Resource):
    def __init__(self, config, rpc_methods):
        self.config = config
        self.dispatcher = Dispatcher()
        for m in rpc_methods:
            self.dispatcher.add_method(m)

class KMEListener(BaseJRPCListener):
    def __init__(self, config):
        rpc_methods = [ RegisterWorkorderProcessor,                                   
                                    PreporocessWorkorder]
        super().__init__(config, rpc_methods)      
```

common/python/listener/base_jrpc_listener.py
```diff
@@ -0,0 +1,224 @@
...
+                jrpc_id,
+                "UNKNOWN_ERROR: unknown exception processing http " +
+                "request {0}: {1}".format(request.path, str(err)))
+            return response
+
+# -----------------------------------------------------------------
+# -----------------------------------------------------------------
+
+
+def start_listener(config, host_name, port, listener):
```
manojgop, line 170: Is config parameter required. It's not used

common/python/listener/base_jrpc_listener.py
```diff
@@ -0,0 +1,224 @@
...
+    root = listener
+    site = server.Site(root)
+    reactor.listenTCP(port, site, interface=host_name)
+
+    logger.info('%s started on port %s', type(listener).__name__, port)
+
+    try:
+        reactor.run()
+    except reactor_error.ReactorNotRunning:
+        logger.warn('shutdown')
```
manju956, line 181: Can we error logs to catch exception?

rranjan3: Makes sense @manju956 . Done.

PR #420, opened 2020-04-16
Bug fix to fabric and ethereum connector

Ram-srini:
1. Submit work order only if active worker id exists in
    the work order request from blockchain event.
 2. Condition to handle synchronous work order processing response

### DIFF
--- a/blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
+++ b/blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
@@ -62,6 +62,8 @@ class FabricConnector():
         self.__fabric_work_order = FabricWorkOrderImpl(self.__config)
         self.__jrpc_worker = JRPCWorkerRegistryImpl(self.__config)
         self.__jrpc_work_order = JRPCWorkOrderImpl(self.__config)
+        # List of active available worker ids in Avalon
+        self.__worker_ids = []
         # Wait time in sec
         self.WAIT_TIME = 31536000
         nest_asyncio.apply()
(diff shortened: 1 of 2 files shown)


blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
```diff
@@ -122,50 +124,57 @@ def workorder_event_handler_func(self, event, block_num, txn_id, status):
...
-            # get the work order result
-            jrpc_req_id += 1
-            work_order_result = \
-                self.__jrpc_work_order.work_order_get_result(
-                    work_order_req['workOrderId'],
-                    jrpc_req_id
-                )
-            logging.info("Work order get result {}".format(
-                work_order_result
+        # Submit the work order to listener if worker id the event
```
danintel, line 127: Change "worker id the event" to "worker id from the event"

blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
```diff
@@ -122,50 +124,57 @@ def workorder_event_handler_func(self, event, block_num, txn_id, status):
...
+        # matches with available worker ids
+        if work_order_req['workerId'] in self.__worker_ids:
```
manojgop, line 129: 1) Is this issue applicable for ethereum as well ? Then we need to do similar change in ethereum.

2) If worker id in request does match available worker ids in Avalon KV storage can we return error.

blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
```diff
@@ -122,50 +124,57 @@ def workorder_event_handler_func(self, event, block_num, txn_id, status):
...
+                        work_order_req['workOrderId'],
+                        jrpc_req_id
+                    )
+                logging.info("Work order get result {}".format(
+                    work_order_result
+                ))
+            # With ZMQ based work order handler in listener returns result
+            # as a response for work order submit
+            elif response and 'result' in response:
+                work_order_result = response
```
manojgop, line 157: Please add similar change for ethereum. We could rephrase the comment "ZMQ based work order handler " to " With Synchronous work order processing" . ZMQ could be replaced with other approach for synchronous work order processing later.

Ram-srini: Currently ethereum is not validating work order submit response. I can add similar work order submit response validation in ethereum as well.

blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
Ram-srini: 1. Same changes required in thereum and i will do it.
2. check is there in https://github.com/hyperledger/avalon/blob/master/listener/avalon_listener/tcs_work_order_handler.py#L121
not in tcs_work_order_handler_sync.py we can similar check here also.

blockchain_connector/avalon_blockchain_connector/fabric/fabric_connector.py
Ram-srini: Done

PR #423, opened 2020-04-18
Update FAQ

danintel:
- Add answer on Avalon nodes
- Add Singleton, KME, and WPE to glossary
- Add workload ID and worker pool


docs/FAQ.rst
```diff
@@ -526,6 +536,9 @@ KECCACK-256
     3", before Keccak was standardized to SHA-3 (FIPS-202) and Keccak has
     minor variations from SHA-3
 
+KME
+    Key Management Enclave Avalon worker. Compare to KME and WPE
```
EugeneYYY, line 540: 1) It should be "Compare with _**Singleton**_ and WPE" 
2) I would add (before the phrase above) "KME is a part of the worker pool responsible for the key management. It has access to the worker's private keys and controls workorder execution by WPE"

docs/FAQ.rst
```diff
@@ -722,6 +739,9 @@ Work order response
     A response generated by an Avalon worker to a work
     order. The response is a JSON string
 
+WPE
+    Work order Processor Enclave Avalon worker. Compare with KME and WPE
```
EugeneYYY, line 743: It should be "Compare with KME and _**Singleton**_"
I would add (before the phrase above) "WPE is a part of the worker pool responsible for the workorder processing. It doesn't has access to the worker's private keys and relies on the KME to provide necessary context for the workorder processing."

docs/FAQ.rst
```diff
@@ -722,6 +739,9 @@ Work order response
...
+
```
EugeneYYY, line 744: I would add worker pool definition also

danintel: Added. I also added workload ID.

danintel:
Hi Eugene,
I made your suggested changes—please re-review. I also added definitions for workload ID and worker pool.

Dan

EugeneYYY:
Done. Nice addition to the FAQ!



Regards,
Eugene

PR #425, opened 2020-04-20
Remove timer based stop event in fabric work_order_get_result API

Ram-srini:

### DIFF
--- a/sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
+++ b/sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
@@ -93,7 +93,9 @@ class EventListener(base.ClientBase):
 
     async def start_event_handling(self):
         """
-        Start event listener.
+        Start event listener and listen for events forever
+        Only way to stop the event listener by calling
+        stop_event_listener()
         """
         def _event_handler(event, block_num, txnid, status):
             logger.debug(
@@ -117,10 +119,46 @@ class EventListener(base.ClientBase):
         finally:
             self._save_config()
 
-    async def stop_event_handling(self, seconds):
+    async def stop_event_handling(self, seconds=0):
         """
         Stop event listener.
         """
         await asyncio.sleep(seconds)
         self._channel_event_hub.unregisterChaincodeEvent(self._event_regid)
         self._channel_event_hub.disconnect()
+
+    async def get_single_event(self):
+        """
+        Start event listener and listen for particular event.
+        Once we got the particular event then unregister and
+        close the event listenerhub.
+        """
+        def _event_handler(event, block_num, txnid, status):
+            logger.debug(
+                'Event: {0}\nblock_num: {1}\ntxid: {2}\nstatus: {3}'.format(
+                    event, block_num, txnid, status))
+            try:
+                # When to stop event handling?
+                # Based the flag return by the _handler
+                is_done = self._handler(event, block_num, txnid, status)
+                if is_done:
+                    asyncio.get_event_loop().run_until_complete(
+                        self.stop_event_handling())
+                self._last_block = block_num
+            except Exception as ex:
+                logger.error('Handler error: {0}'.format(ex))
+
+        self._event_regid = self._channel_event_hub.registerChaincodeEvent(
+            self._chaincode, self._event, start=self._last_block,
+            onEvent=_event_handler)
+        logger.info('Event handler registered!')
+        try:
+            # wait for the connect() to get the stream of events
+            # it will call the event handler function(_event_handler)
+            # it will be unblocked only when call disconnect()
+            await self._channel_event_hub.connect(False)
+        except grpc.RpcError as ex:
+            # This is expected when event hub gets disconnected
+            logger.info("Event handler disconnected")
+        finally:
+            self._save_config()
--- a/sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
+++ b/sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
@@ -18,6 +18,7 @@ import logging
 from os import environ
 import time
 import asyncio
+import nest_asyncio
 
 from utility.hex_utils import is_valid_hex_str
 from avalon_sdk.connector.blockchains.common.contract_response \
@@ -48,8 +49,7 @@ class FabricWorkOrderImpl(WorkOrderProxy):
         self.CHAIN_CODE = 'order'
         self.WORK_ORDER_SUBMITTED_EVENT_NAME = 'workOrderSubmitted'
         self.WORK_ORDER_COMPLETED_EVENT_NAME = 'workOrderCompleted'
-        self.WAIT_TIME = 30
-        self.__wo_resp = ''
+        nest_asyncio.apply()
         if config is not None:
             self.__fabric_wrapper = FabricWrapper(config)
         else:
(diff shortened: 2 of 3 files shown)


sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
```diff
@@ -119,33 +119,35 @@ def handle_fabric_event(event, block_num, txn_id, status):
             payload = event['payload'].decode("utf-8")
             resp = json.loads(payload)
             response = json.loads(resp["workOrderResponse"])
+            # to use global scope of wo_resp
+            nonlocal wo_resp
```
manojgop, line 123: Can't we directly use the response in events payload ?. Any specific reason why we need to use another variable with global scope ?  How are the changes in PR aligned with overall design of making event handling common across Fabric and Ethereum ?

Ram-srini: Function handle_fabric_event() nested function inside the work_order_get_result(), we can't return result from there because this handler function called from other places. we have to use variable in the outer function, and update the variable in the nested function. These changes are specific to work_order_get result().

To make event handling common across Fabric and Ethereum, we have to come up with apis in the interface to create event handler.

sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
```diff
@@ -124,3 +126,36 @@ async def stop_event_handling(self, seconds):
...
+                self._last_block = block_num
+            except Exception as ex:
+                logger.error('Handler error: {0}'.format(ex))
+
+        self._event_regid = self._channel_event_hub.registerChaincodeEvent(
+            self._chaincode, self._event, start=self._last_block,
+            onEvent=_event_handler)
+        logger.info('Event handler registered!')
+        try:
+            await self._channel_event_hub.connect(False)
```
manojgop, line 156: When does this function gets unblocked. Is it when disconnect() is called from event_handler. Not very obvious. Could you please add a comment in code.

sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
manojgop: if `handle_fabric_event()` is called from other places then why do we need to make this function nested since the scope of the function is not limited to `work_order_get_result()`. Also `wo_resp` variable in `handle_fabric_event()` makes sense only if it's used within `work_order_get_result()`.  

If you plan to modify it as part of unifying event handling for Fabric and Ethereum then  separate PR can be raised to handle this as well

Ram-srini: Scope of the function handle_fabric_event() is nested within the work_order_get_result(). Instead of defining the function out side, within the class it is good to have nested function. Do you see any difficulties in that? To know more about nested function go through this - https://stackabuse.com/python-nested-functions
Variable wo_resp variable is used in the work_order_get_result() it local to both the functions.  

This how fabric events handled with callback handler function. This change is not unify event handling approach.

Ram-srini: This approach we can't unify it because we created separate implementation class to handle it based the sdk what is provided the blockchain.

sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
Ram-srini: connect is async call please do refer the fabric-sdk-python for information.

sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
manojgop: I'm fine with nested function here as long as scope of nested function is within outer function `work_order_get_result()`. Is that the case now ? I'm confused with your above comment ---.> "because this handler function called from other places"

manojgop: Are we planning to raise a separate PR define common interface for event handling.  Specific blockchain SDK API calls will be implementation specific. But can't we make asyncio constructs for handling events common?.

sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
manojgop: ok. It was not obvious to me looking at the code how this function is waiting to get a single event. I couldn't find Fabric Python SDK doc. Instead I went through the documentation in this link https://hyperledger.github.io/fabric-sdk-node/release-1.4/ChannelEventHub.html 

According to this `registerChainCodeEvent()` should be called after calling `connect()`.  This is what `connect` documentation mentions -> "Register event listeners and the error callbacks by calling any one of the registerBlockEvent, registerTxEvent or registerChaincodeEvent methods, after calling connect()"

sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
Ram-srini: > I'm fine with nested function here as long as scope of nested function is within outer function `work_order_get_result()`. Is that the case now ? I'm confused with your above comment ---.> "because this handler function called from other places"

variable scope will be within the function work_order_get_result(). I meant Handler function will be called once we get the event, we are not calling it explicitly.

sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
Ram-srini: That is fabric node sdk documentation. Refer the sample code here - https://github.com/hyperledger/fabric-sdk-py/blob/master/test/integration/e2e_test.py

litong01:
I am not really sure where this patch is going, removal of the time limit is fine, but I would suggest to have few functions like this `async def stop_event_handling(self, seconds):`, the one current we have there is to stop listening after a period of time, but you can create few more of similar functions on different conditions to stop listening the events. For example, you can create another function (or adding logic to the existing one) and expect the 2nd parameter not to be an integer but an object which describes particular condition, when the condition satisfies, then stop the listening process. With python, you can even pass a function there to do that. The code we have there is already very generic in terms of returning the actual events. Both stopping condition(time currently, more conditions need to be added) and handling callbacks are already there. I do not think there is more than that need to be done unless I misunderstood the requirement.

Ram-srini:
> I am not really sure where this patch is going, removal of the time limit is fine, but I would suggest to have few functions like this `async def stop_event_handling(self, seconds):`, the one current we have there is to stop listening after a period of time, but you can create few more of similar functions on different conditions to stop listening the events. For example, you can create another function (or adding logic to the existing one) and expect the 2nd parameter not to be an integer but an object which describes particular condition, when the condition satisfies, then stop the listening process. With python, you can even pass a function there to do that. The code we have there is already very generic in terms of returning the actual events. Both stopping condition(time currently, more conditions need to be added) and handling callbacks are already there. I do not think there is more than that need to be done unless I misunderstood the requirement.

@litong01 This specific requirement is to get the single event from eventhub and close it after getting the event. We are using it in work_order_get_result. Once we get the workOrderCompleted event then we are closing eventhub and unregister an Chaincodeevnet. May be we can call stop_event_handling() in get_single_event()

litong01:
@Ram-srini in that case, I would suggest to simply call stop_event_handling with 0 as the 2nd parameter in your event handler.

sdk/avalon_sdk/connector/blockchains/fabric/event_listener.py
```diff
@@ -93,7 +93,9 @@ def _save_config(self):
 
     async def start_event_handling(self):
         """
-        Start event listener.
+        Start event listener and listen for events forever
+        Only way to stop the event listener by calling
```
manojgop, line 97: by -> is by

PR #426, opened 2020-04-23
Add C++ Header Files to Avalon Reference Manual

danintel:
C/C++ source files need a @file block comment to be included
in Doxygen document generation. Otherwise the file will be ignored.

### DIFF
--- a/common/cpp/crypto/crypto_utils.h
+++ b/common/cpp/crypto/crypto_utils.h
@@ -13,6 +13,13 @@
  * limitations under the License.
  */
 
+/**
+ * @file
+ * Avalon Crypto Utilities:
+ * hashing, base 64 conversion, random number generation, key generation,
+ * encrypt, and decrypt.
+ */
+
 #pragma once
 
 #include <vector>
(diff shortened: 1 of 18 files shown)


common/cpp/crypto/crypto_utils.h
```diff
@@ -13,6 +13,13 @@
  * limitations under the License.
  */
 
+/**
+ * @file
+ * Avalon Crypto Utilities:
+ * hashing, base 64 conversion, random number generation, key generation,
+ * encrypt, and decrypt.
```
manojgop, line 20: Do we want to add crypto C++ header file in Avalon Reference Manual because, we expect Avalon users to use these C++ crypto function in C++ SGX workloads ?

danintel: Yes. It is being used in simple wallet now, for example.

manojgop: There are other crypto functionalities like RSA or ECDSA which can be used by SGX workloads. Are we planning to include those too in documentation ?

danintel: Maybe later. Public key encryption seems to be spread out among 6 header files and seems to be hard-coded to specific algorithms and keylengths that are currently used internally:

pkenc.h              pkenc_public_key.h  sig_private_key.h
pkenc_private_key.h  sig.h               sig_public_key.h

PR #427, opened 2020-04-24
Cleanup Base64 Decoding

danintel:
In verify_signature(), replace call to EVP_DecodeBlock_wrapper()
with EVP_DecodeBlock.

Remove redundant and unused functions
decode_base64_block() and EVP_DecodeBlock_wrapper().

### DIFF
--- a/common/cpp/crypto/verify_signature.cpp
+++ b/common/cpp/crypto/verify_signature.cpp
@@ -18,16 +18,17 @@
 #include <string.h>
 #include <openssl/x509v3.h>
 #include <openssl/pem.h>
+#include <openssl/evp.h>
 
 #include "verify_signature.h"
 #include "c11_support.h"
 #include "crypto_utils.h"
 
 bool verify_signature(const char* cert_pem,
-                                 const char* msg,
-                                 unsigned int msg_len,
-                                 char* signature,
-                                 unsigned int signature_len)
+                      const char* msg,
+                      unsigned int msg_len,
+                      char* signature,
+                      unsigned int signature_len)
 {
     X509* crt = NULL;
     int ret;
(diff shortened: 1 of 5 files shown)


common/cpp/crypto/verify_signature.cpp
```diff
@@ -18,16 +18,17 @@
 #include <string.h>
 #include <openssl/x509v3.h>
 #include <openssl/pem.h>
+#include <openssl/evp.h>
```
manju956, line 21: This was not included in crypto_utils.cpp before. Is inclusion of this header file needed?

common/cpp/crypto/verify_signature.cpp
```diff
@@ -47,13 +48,14 @@ bool verify_signature(const char* cert_pem,
...
-                                  (unsigned char*)signature,
-                                  signature_len);
-    assert(ret!=-1);
+    // Decode base64-encoded signature.
+    // Decoded output may have 1-3 extra '\0' bytes appended.
+    static const int signature_decoded_len = 2048;
+    unsigned char signature_decoded[signature_decoded_len + 3];
+    ret = EVP_DecodeBlock(signature_decoded,
+                          (unsigned char*)signature,
+                          signature_len + 3);
```
manju956, line 57: Extra buffer size to allocate \o characters was not there when we are using the EVP_DecodeBlock_wrapper() function. Was that a miss?

danintel:
Merged in PR #429

common/cpp/crypto/verify_signature.cpp
danintel: No. That was avoided by having EVP_DecodeBlock_wrapper(), which I removed. The only function it did was trim the extra \0 characters.

manju956: Ok.

PR #430, opened 2020-04-28
Refactor enclave helper/bridge

rranjan3:
- Rename avalon_enclave_bridge to avalon_enclave_info
- Move EnclaveHelper class from avalon_enclave_help to
  avalon_enclave_info
- Rename EnclaveHelper to EnclaveInfo
- Update enclave_manager for refactor


enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -267,3 +274,79 @@ def create_signup_info(nonce):
 
     # Now we can return the real object
     return signup_info_obj
+
+
+# -----------------------------------------------------------------
+class EnclaveInfo(object):
+    """
+    Wraps calls to the client for symmetry with the enclave service client
```
manju956, line 282: Can we rephrase the function comment?

enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -267,3 +274,70 @@ def create_signup_info(nonce):
...
+    """
+
+    # -------------------------------------------------------
+    @classmethod
+    def create_enclave_signup_data(cls):
+        """
+        Create enclave signup data
+        """
+
+        nonce = '{0:016X}'.format(random.getrandbits(64))
```
danintel, line 293: This should be 128 bits (16 bytes) instead of 64 bits. The nonce is never converted to a integer, so there is no overflow problem and avoids the birthday paradox.  That is change this to:
`nonce = '{0:032X}'.format(random.getrandbits(128))`

enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -98,7 +97,15 @@ def __update_sig_rl():
 
 
 # -----------------------------------------------------------------
-def initialize_with_configuration(config):
+def initialize_enclave(enclave_config):
+    """initialize_enclave -- Call the initialization function on the
+    enclave module
+    """
+    return _initialize_with_configuration(enclave_config)
```
manojgop, line 104: Can we move the function `_initialize_with_configuration()` body here and delete `_initialize_with_configuration()`

enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -267,3 +274,70 @@ def create_signup_info(nonce):
 
     # Now we can return the real object
     return signup_info_obj
+
+
+# -----------------------------------------------------------------
+class EnclaveInfo(object):
+    """
```
manojgop, line 281: Can we move all the functions inside `EnclaveInfo()` class. May be rename this class to `EnclaveSignupInfo()`

enclave_manager/avalon_enclave_manager/singleton/enclave_manager.py
```diff
@@ -347,13 +347,13 @@ def __update_receipt(self, kv_helper, wo_id, wo_json_resp):
 
 
 # -----------------------------------------------------------------
-def create_enclave_signup_data():
+def _create_enclave_signup_data():
```
manojgop, line 350: Is this function required ? Can't we directly call `enclave_info.EnclaveInfo.create_enclave_signup_data()`

enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -0,0 +1,339 @@
+# Copyright 2018 Intel Corporation
```
manojgop, line 1: 2018 -> 2020

enclave_manager/avalon_enclave_manager/avalon_enclave_info.py
```diff
@@ -0,0 +1,339 @@
...
+        except KeyError as ke:
+            raise Exception("missing enclave initialization parameter; {}"
+                            .format(str(ke)))
+
+        self.enclave_keys = \
+            keys.EnclaveKeys(self.verifying_key, self.encryption_key)
+
+    # -------------------------------------------------------
+
+    def create_enclave_signup_data(self):
```
manojgop, line 75: Looks like this function is not used outside this class. If so, this is a private function and as a convention please rename this to `__create_enclave_signup_data()`. Similarly pre-pend single/double underscore for other functions in this class which can be private member functions. 
https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

PR #431, opened 2020-04-29
Segregate common trusted signup APIs

manju956:
Common signup related APIs between different type of workers are added
to separate file.

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave/signup_enclave.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave/signup_enclave.cpp
@@ -46,10 +46,6 @@
 #include "signup_enclave_util.h"
 
 
-// Initializing singleton class object which gets initialized when
-// getInstance is called
-EnclaveData* EnclaveData::instance = 0;
-
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 // XX Declaration of static helper functions                         XX
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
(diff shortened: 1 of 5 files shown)


tc/sgx/trusted_worker_manager/enclave/signup_enclave_common.h
```diff
@@ -0,0 +1,28 @@
...
+
+tcf_err_t ecall_CalculateSealedEnclaveDataSize(size_t* pSealedEnclaveDataSize);
+
+tcf_err_t ecall_CalculatePublicEnclaveDataSize(size_t* pPublicEnclaveDataSize);
+
+tcf_err_t ecall_CreateEnclaveData(size_t* outPublicEnclaveDataSize,
+    size_t* outSealedEnclaveDataSize);
+
+tcf_err_t ecall_UnsealEnclaveData(char* outPublicEnclaveData,
+    size_t inAllocatedPublicEnclaveDataSize);
```
manojgop, line 28: Not sure why is this called common. Sealing is not applicable for WPE. Do you mean common for Singleton and KME. May be call it signup_seal.h ?

Creating C++ class might be better option than functional coding approach. But if changes are too much to do in this sprint, that can be done later.

manju956: I named this file as signup_enclave_common because all functions except UnsealEnclaveData are common. the function UnsealEnclaveData can be stopped from calling even at the untrusted layer itself by not providing the wrapper function for the same function (for WPE).

I thought about creating class in trusted code, but i am not sure about any way to instantiate class in trusted layer from untrusted and call actual ecalls.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_common.cpp
```diff
@@ -0,0 +1,148 @@
...
+            TCF_LOG_ERROR, "Unknown error in Avalon enclave(ecall_CalculateSealedEnclaveDataSize)");
+        result = TCF_ERR_UNKNOWN;
+    }
+
+    return result;
+}  // ecall_CalculateSealedEnclaveDataSize
+
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+tcf_err_t ecall_CalculatePublicEnclaveDataSize(size_t* pPublicEnclaveDataSize) {
+    tcf_err_t result = TCF_SUCCESS;
```
rranjan3, line 56: Preferable to start initialize with failure.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_kme.cpp
```diff
@@ -38,11 +38,11 @@
 
 #define KME_SIGNUP_EXT_DATA_SIZE 32
 
-void CreateReportDataKME(std::string& enclave_signing_key,
+static void CreateReportDataKME(std::string& enclave_signing_key,
     const uint8_t* ext_data,
     sgx_report_data_t* report_data);
 
-void CreateSignupReportDataKME(const uint8_t* ext_data,
+static void CreateSignupReportDataKME(const uint8_t* ext_data,
```
rranjan3, line 45: Any specific purpose to switch to static?

tc/sgx/trusted_worker_manager/enclave/signup_enclave.cpp
```diff
@@ -218,46 +132,6 @@ tcf_err_t ecall_CreateSignupData(const sgx_target_info_t* inTargetInfo,
     return result;
 }  // ecall_CreateSignupData
 
-// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-tcf_err_t ecall_UnsealEnclaveData(
```
rranjan3, line 222: Removing these functions from here should have caused breakage. I do not see the refactored files where these functions were being used.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_common.cpp
manju956: Agree, this needs to be followed in all CPP functions. I will make the changes here and rest of the places can be handled as and when we touch the code.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_kme.cpp
manju956: Since the function is local to the scope of the file, its made static.

tc/sgx/trusted_worker_manager/enclave/signup_enclave.cpp
manju956: It should not break, because SGX utils will automatically route the API calls to the location where they are defined.

PR #432, opened 2020-04-30
Fix Buffer Overrun in Base64 Decoding

danintel:
Output, but not input needs +3 bytes.
Regression from PR #429.


common/cpp/crypto/verify_signature.cpp
```diff
@@ -54,7 +54,7 @@ bool verify_signature(const char* cert_pem,
     unsigned char signature_decoded[signature_decoded_len + 3];
     ret = EVP_DecodeBlock(signature_decoded,
                           (unsigned char*)signature,
-                          signature_len + 3);
+                          signature_len3);
```
manju956, line 57: change signature_len3 to signature_len

Issue #101, opened 2019-09-27
Encryption RSA key is "compact"


kubasiemion:
The encryption RSA key is generated and marshalled to a "pem" format at the TCS start:
`JSON serialized worker info is {"workerType": [...], "encryptionKey": "-----BEGIN RS
A PUBLIC KEY-----\nMIIBCgKCAQEAsq7Bbdkb/O7Z05zAqONVteRZC/G1WgwiTpNuscUjnlLs9yqvux3x\n6GJESSeQ7kEtz/lPcdqkWHXPAjpFh1HPqoqklkQr7tjCzwYgC9hA8L6byJPaErdC\nDJXmJkHwiBMDuRPTMGOBQuBIp74kf6w/1EhfYjgpgyhpZVHCk8aPfaydG8PgNwGu\niJCLvLBEXRfNSmDBTGgFSdMMp+owGF0xDbd8bgG3Mk3tEpbQMaDXjsJkQ1mn3P0s\nw1Znx8S9u9r0nSX6xFl2mKqAODEBdAf14pjmbOe9U0DEWHE/ZjaeNsgHXfVB/st2\nXeTYgiW2iheqnZ0SYmqGbqasB++3Sq0IiwIDAQAB\n-----END RSA PUBLIC KEY-----\n", "encryptionKeySignature": ""}}, "status": 1}` 
This is the format in which it is being subsequently propagated.
Alas, one cannot easily parse this key (try to paste into a *.pem file and use as input for openssl). That is because only carries the internal part of the key (270 bytes). It misses the RSA OID, which, together with ASN1/DER formatting would make it 294 bytes long. 
Now I may not know what exactly I am writting about, but it took me a hack to get the interoperability with other std crypto libraries (go/crypto, go/spacemonkeys, ...). A hack that is not available on StackExchange! So how about our young audience?

danintel:
The key should not be hard-coded to a specific algorithm.  Including the RSA OID will allow changes in the future and make it more flexible.

kubasiemion:
Well, it is more of a hunch how it should be written rather than a "suggested fix" :-)
But yes, to my uninitiated eye that is the way to go. Very good point on the flexibility of the code -  honestly, have not even thought about it... 
I can work out a full fix, but probably only after Devcon. Unless somebody takes care of that earlier.

> The root reason (I think) is that in
> `trusted-compute-framework/tc/sgx/common/crypto/pkenc_public_key.cpp`
> the methods
[...]

PR #434, opened 2020-04-30
Change the RSA Public Key serialization to FC2437

kubasiemion:
I recommend a small change in the code to make the "encryptionKey"  json field comply with PKCS 1. This is achieved by using PEM_..._bio_RSA_PUBKEY methods instead of PEM_..._bio_RSAPublicKey ones. 
This would make interop with Java, Go, etc., actually working.
Currently the encoded format is something like:
0  135: SEQUENCE {
   3  129:   INTEGER
         :     00 C6 E9 382 [...]
 135    1:   INTEGER 3
         :   }

The new structure (with the standard RSA OID) would be:

0  157: SEQUENCE {
   3   13:   SEQUENCE {
   5    9:     OBJECT IDENTIFIER rsaEncryption (1 2 840 113549 1 1 1)
  16    0:     NULL
         :     }
  18  139:   BIT STRING, encapsulates {
  22  135:     SEQUENCE {
  25  129:       INTEGER
         :         00 C7 6F [..] 
   157    1:       INTEGER 3
         :       }
         :     }
         :   }

### DIFF
--- a/common/cpp/crypto/pkenc_public_key.cpp
+++ b/common/cpp/crypto/pkenc_public_key.cpp
@@ -59,7 +59,7 @@ RSA* deserializeRSAPublicKey(const std::string& encoded) {
         throw Error::RuntimeError(msg);
     }
 
-    RSA* public_key = PEM_read_bio_RSAPublicKey(bio.get(), NULL, NULL, NULL);
+    RSA* public_key = PEM_read_bio_RSA_PUBKEY(bio.get(), NULL, NULL, NULL);
     if (!public_key) {
         std::string msg(
             "Crypto Error (deserializeRSAPublicKey): Could not "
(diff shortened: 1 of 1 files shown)


kubasiemion:
This is solving issue #101

danintel:
@kubasiemion ,
What is a FC2437?

I can integrate this when you sign your commit.

The current implementation uses  PEM_read_bio_RSAPublicKey(), which reads PKCS#1 format.
This expects a RSAPublicKey structure, which starts with `-----BEGIN RSA PUBLIC KEY-----`.
Generate this with `openssl rsa -in key.pri -RSAPublicKey_out -out key.pub2`

The change uses PEM_read_bio_RSA_PUBKEY(), which expects a SubjectPublicKeyInfo structure.
This  starts with `-----BEGIN PUBLIC KEY----`.
Generate this with `openssl rsa -in key.pri -pubout -out key.pub1`

Ref:
https://stackoverflow.com/questions/7818117/why-i-cant-read-openssl-generated-rsa-pub-key-with-pem-read-rsapublickey

danintel:
@kubasiemion,
I took your ideas and used them in PR#502.
The latter PR accepts both format strings for deserialization (BEGIN RSA PUBLIC KEY and BEGIN PUBLIC KEY),
but only writes BEGIN PUBLIC KEY pem strings.
I also add some tests to avoid future regression.

danintel:
Since this PR is unsigned I cannot use it due to legal requirements by The Linux Foundation.
I implemented the same functionality in PR #502, which uses your ideas, and is now integrated into Avalon. 
So this PR is now moot and I am closing it.

PR #435, opened 2020-04-30
Documentation Fixes

danintel:
- docs/README.md: Doxygen breaks graphic link in .md file after processing
- docs/README.md: Add link to SDK Reference Manual
- docs/FAQ.rst: Add answer about ``Invalid interpolation format`` error
- .gitattributes: Add .rst to treat rst markup files as text files

### DIFF
--- a/common/cpp/crypto/verify_signature.cpp
+++ b/common/cpp/crypto/verify_signature.cpp
@@ -24,6 +24,16 @@
 #include "c11_support.h"
 #include "crypto_utils.h"
 
+/**
+ * Verify that cert_pem is signed by CA,
+ * using CA certificate ca_cert_pem as a root of trust.
+ *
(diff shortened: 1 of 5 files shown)


common/cpp/crypto/verify_signature.cpp
```diff
@@ -54,7 +54,7 @@ bool verify_signature(const char* cert_pem,
     unsigned char signature_decoded[signature_decoded_len + 3];
     ret = EVP_DecodeBlock(signature_decoded,
                           (unsigned char*)signature,
-                          signature_len + 3);
+                          signature_len);
```
manju956, line 57: Can you remove this file from commit as this is not documentation fix?

danintel: That was  rebase error. That fix is already in the code.  I fixed it--please re-review.

PR #436, opened 2020-05-01
Certificate and Signature Validation Fixes

danintel:
verify_signature()
- Fix regression caused by length too long when '=' padding characters present
- Check buffer is large enough before calling EVP_DecodeBlock()

For both:
verify_certificate_chain() and verify_signature()
- Add function interface comments for Doxygen
- Add sanity checks
- Clean up return code

### DIFF
--- a/common/cpp/crypto/verify_signature.cpp
+++ b/common/cpp/crypto/verify_signature.cpp
@@ -24,28 +24,67 @@
 #include "c11_support.h"
 #include "crypto_utils.h"
 
+
+/*
+ * EVP_DecodeBlock pads its output with \0 if the output length is not
+ * a multiple of 3. Check if the base64 string is padded at the end with '='
+ * and adjust the output length.
+ * For details, see the OpenSSL EVP_EncodeInit man page.
+ */
+#define ADJUST_DECODE_LENGTH(in, in_len) \
+    ((((in)[(in_len) - 1] == '=') ? -1 : 0) + \
+     (((in)[(in_len) - 2] == '=') ? -1 : 0))
+
+
+/* Calculate the length of a base 64 encoded string after decoding. */
+static unsigned int
+base64_decoded_length(const char *encoded, unsigned int encoded_len)
+{
+    if (encoded == NULL || encoded_len == 0) {
+        return 0;
+    }
+
+    unsigned int len = (encoded_len / 4) * 3;
+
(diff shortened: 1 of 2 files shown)


common/cpp/crypto/verify_signature.cpp
```diff
@@ -24,18 +24,65 @@
...
 bool verify_signature(const char* cert_pem,
-                      const char* msg,
-                      unsigned int msg_len,
-                      char* signature,
-                      unsigned int signature_len)
+    const char* msg, unsigned int msg_len,
+    const char* signature, unsigned int signature_len)
 {
-    X509* crt = NULL;
+    static const unsigned int max_decoded_len = 2048; // <=16384 bit signatures
```
manojgop, line 71: So decoded signature length can't be greater than 2048 bytes ? Is it based on a specific signature algorithm rules ? Could you please add a comment what does 16384 bit signatures mean in above comment?

manojgop: Does it make sense to use the function  `EVP_PKEY_size()` https://www.openssl.org/docs/man1.1.0/man3/EVP_PKEY_size.html

danintel: @manojgop,
Keys more than 2048 bytes (16384) are not in practical use because they take weeks or months to generate.
 I added a comment about that.  This is existing code and I did not change it.

The bigger problem is malicious or incorrect code can create a buffer overflow by passing a real or fake key that is >16384.

No, `EVP_PKEY_size()` is the wrong function. It is the maximum key size of an existing, decoded key.

The function here `base64_decoded_length()` computes the exact decoded length of a base64 encoded buffer.

danintel:
@manjop I wrote test programs to test this and other crypto functions and they work OK.
I will integrate the test programs separately.

manju956:
> @manju956 I wrote test programs to test this and other crypto functions and they work OK.
> I will integrate the test programs separately.

Ok. Please test Avalon in HW mode once to ensure sanity.

manju956:
> > @manju956 I wrote test programs to test this and other crypto functions and they work OK.
> > I will integrate the test programs separately.
> 
> Ok. Please test Avalon in HW mode once to ensure sanity.

I checked out your PR changes locally and tested SGX HW, and the flow works fine.

Issue #433, opened 2020-04-30
RSA encryption key size inconsistent (code/docs)


kubasiemion:
This is in common/cpp/crypto:
In the README.md: 

> Asymmetric encryption | RSA-OAEP | 3072 | (1)

In the pkenc.h:

>  //***RSA is not quantum resistant ***//
> //*** USE 3072 for long term security ***//
> const int RSA_KEY_SIZE = 2048;

I understand this is a transition-issue, but still confusing...

danintel:
Yes, this needs to be updated to RSA-3072.

danintel:
Not a bug.
Withdrawing this PR. The code is correct.  The 2048 byte buffer allows up to RSA-16384 (2048*8) bits.
I added a comment with PR #443 to avoid future confusion.

PR #444, opened 2020-05-05
Change RSA-2048 to RSA-3072

danintel:
Change RSA key size from 2048 to 3072 bits.
This is as documented in common/cpp/crypto/README:

> Cryptographic Primitives Used
> -----------------------------
> | Primitive             | Algorithm | Keysize | Comments |
> | --------------------- | --------- | ------- | -------- |
> | Asymmetric encryption | RSA-OAEP  | 3072    | (1)      |
> (1) Not PQ resistant

Fixes issue #433.


danintel:
Withdrawing this PR. The code is correct.  The 2048 byte buffer allows up to RSA-16384 (2048*8) bits.

PR #450, opened 2020-05-07
Nginx based Load Balancer for multiple http listeners

pankajgoyal2:
This PR demostrates
- how a Load Balancer (nginx) can be used to forward the
  client requests to multiple listeners in backend to be able to
  handle multiple requests.

### DIFF
--- a/docker/lb/README.md
+++ b/docker/lb/README.md
@@ -1,0 +1,28 @@
+# nginx based Load Balancer for Avalon
+
+This shows how to create nginx based Load Balancer to forward the requests
+coming from clients to one of the many avalon-listeners running in backend.
+
+## Running
+
+1. Build and run Avalon components.
+    ```bash
+    docker-compose -f docker/lb/docker-compose-lb.yaml up --build
+
+    The above command is going to start nginx based service (avalon-lb) along
+    with other avalon services. The nginx load balancer configuration is
+    defined in nginx.conf file which starts the nginx server on port 9947.
+    The client is going to submit the transaction to avalon-lb Load Balancer
+    and this load balancer is going to forward the  requests to 3 instances
+    of avalon-listeners in the backend in round-robin fashion.
+    ```
+
+2. Start transaction from client.
+    ```bash
+    # Go to shell container
+    docker exec -it avalon-shell bash
+ 
+    # run the transaction
+    cd examples/apps/generic_client
+    ./generic_client.py --uri "http://avalon-lb:9947" --workload_id "echo-result" --in_data "Hello"
+     ```
--- a/docker/lb/docker-compose-lb.yaml
+++ b/docker/lb/docker-compose-lb.yaml
@@ -1,0 +1,201 @@
+# Copyright 2020 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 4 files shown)


docker/lb/README.md
```diff
@@ -0,0 +1,27 @@
...
+    ```
+
+2. Start transaction from client.
+    ```
+    # Go to shell container
+    docker exec -it avalon-shell bash
+ 
+    # run the transaction
+    cd examples/apps/echo/client
+    ./echo_client.py -m "Hello world" -s "http://avalon-lb" -dh
```
manojgop, line 26: Can we use generic client instead of echo client. Echo client will be deprecated soon and replaced with generic client

docker/lb/docker-compose-lb.yaml
```diff
@@ -0,0 +1,201 @@
+# Copyright 2019 Intel Corporation
```
manojgop, line 1: 2019 -> 2020

docker/lb/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+1. Build and run Avalon components.
+    ```bash
+    docker-compose -f docker/lb/docker-compose-lb.yaml up --build
+
+    The above command is going to start nginx based service (avalon-lb) along
+    with other avalon services. The client is going to submit the transaction
+    to avalon-lb Load Balancer and this load balancer is going to forward the
+    requests to 3 instances of avalon-listeners in the backend in round-robin
+    fashion.
```
danintel, line 16: Add where the load balancing is configured (that is in ngix.conf).
Also add that ngix listens to port 80 and forwards requests to one of the backend servers.
I would also use another port besides 80 for an example, as 80 is commonly used for the web.
For example port 9947
 So change the generic_client.py line to use http://avalon-lb:9947 and change 80 to 9947 in nginx.conf

docker/lb/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+2. Start transaction from client.
+    ```bash
+    # Go to shell container
+    docker exec -it avalon-shell bash
+ 
+    # run the transaction
+    cd examples/apps/generic_client
+    ./generic_client.py --uri "http://avalon-lb" --workload_id "echo-result" --in_data "Hello"
+     ```
```
manojgop, line 28: One good test case for load balancer would be to sent multiple requests to Avalon and check if Avalon listeners is able to handle it concurrently.

docker/lb/docker-compose-lb.yaml
```diff
@@ -0,0 +1,201 @@
...
+      args:
+        - http_proxy
+        - https_proxy
+        - no_proxy
+    environment:
+      - http_proxy
+      - https_proxy
+      - no_proxy
+    expose:
+      - 1947
```
manju956, line 141: Can all listener services use same port to listen to?

pankajgoyal2: Yes. Service names are different!

manju956: Ok.

docker/lb/README.md
```diff
@@ -0,0 +1,28 @@
...
+    ```
+
+2. Start transaction from client.
+    ```bash
+    # Go to shell container
+    docker exec -it avalon-shell bash
+ 
+    # run the transaction
+    cd examples/apps/generic_client
+    ./generic_client.py --uri "http://avalon-lb" --workload_id "echo-result" --in_data "Hello"
```
danintel, line 27: Change to http://avalon-lb:9947

PR #457, opened 2020-05-11
Synchronous work order execution changes

Ram-srini:
listener toml uses zmq_url is set to docker service name of enclave manager
(avalon-enclave-manager) and in hardware mode sgx docker compose file is
different so it was breaking. Changed the docker service name for enclave manager.

### DIFF
--- a/docker-compose-sgx.yaml
+++ b/docker-compose-sgx.yaml
@@ -39,11 +39,11 @@ services:
     entrypoint: bash -c "tail -f /dev/null"
     stop_signal: SIGKILL
     depends_on:
-      - avalon-sgx-enclave-manager
+      - avalon-enclave-manager
       - avalon-listener
 
-  avalon-sgx-enclave-manager:
-    container_name: avalon-sgx-enclave-manager
+  avalon-enclave-manager:
+    container_name: avalon-enclave-manager
     image: avalon-sgx-enclave-manager-dev
     build:
       context: .
@@ -58,6 +58,9 @@ services:
       - http_proxy
       - https_proxy
       - no_proxy
+    expose:
+      # ZMQ socket port.
+      - 5555
     volumes:
       # Map aesm service on host to docker, required only in HW mode.
       - /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
@@ -110,6 +113,8 @@ services:
       - no_proxy
     expose:
       - 1947
+      # ZMQ socket port.
+      - 5555
     command: |
       bash -c "
         avalon_listener --bind http://avalon-listener:1947 --lmdb_url http://avalon-lmdb:9090


docker-compose-sgx.yaml
```diff
@@ -39,12 +39,12 @@ services:
...
     stop_signal: SIGKILL
     depends_on:
-      - avalon-sgx-enclave-manager
+      - avalon-enclave-manager
       - avalon-listener
 
-  avalon-sgx-enclave-manager:
-    container_name: avalon-sgx-enclave-manager
-    image: avalon-sgx-enclave-manager-dev
+  avalon-enclave-manager:
```
manju956, line 45: The reason we named the service this way is to differentiate between enclave manager running in SIM mode and HW mode. In the listener config file, by default we can config zmq_url to point to sim mode enclave manager service.
You can mention in a comment that in case of HW to use HW mode enclave manager service.

Ram-srini: We are sharing dockerfile for both SIM and HW mode then, what are we differentiating between these 2 modes? Are we supporting to run both sim and HW modes at a time in a machine?

manju956: Yes, we currently use same dockerfile for both SIM and HW mode. Differentiating factor is SGX_MODE arg that is passed from compose file. Also, in case of HW mode, we map the device from compose file. Its possible to run both SIM and HW modes on SGX enabled platform but not at a time.

manju956:
> listener toml uses zmq_url is set to docker service name of enclave manager
> (avalon-enclave-manager) and in hardware mode sgx docker compose file is
> different so it was breaking. Changed the docker service name for enclave manager.
> 
> Signed-off-by: Hariharan Subramanian [hariharanx.subramanian@intel.com](mailto:hariharanx.subramanian@intel.com)

Modify signed-off field to your name and email.

docker-compose-sgx.yaml
Ram-srini: If both SIM and HW modes are not running at time then why do we need to need to differentiate with docker service name. I agree SGX_MODE env variable and device mount 
 needs to passed in case of hardware mode, but these run time arguments to docker service to run(volume and ENV variable). I don't any issue in using same docker service name for both SIM and HW mode.

manojgop: In the same system we can run both SGX SIM and HW mode. So we need to have different docker image name for enclave manager one for SGX SIM(avalon-enclave-manager-dev) and SGX HW(avalon-sgx-enclave-manager-dev) .

Ram-srini: @manojgop Docker images for both SGX SIM and HW mode is same because  Dockerfile is same for both modes. All we need run time arguments for HW mode ENV and volume mounting, that is the only difference.

Ram-srini: I changed i to same docker service name in both SIM and HW docker file, it is working for me.

manojgop: Dockerfile is same. But docker image generated is different for SGX HW and SIM. The same image won't work for SGX HW and SIM.  check https://github.com/hyperledger/avalon/blob/master/enclave_manager/Dockerfile#L211

danintel: For configuration in `listener/listener_config.toml` and documentation in `docs/FAQ.rst` it is easier if we have the same hostname for both Intel SGX SIM and HW modes... that is `tcp://avalon-enclave-manager:5555`

manojgop: @danintel @Ram-srini  We can probably keep the service name same , but let's keep different docker image name for SGX HW and SIM. If we have same docker image name then on same system it's not easier to test both since if we build image for SGX-SIM and then for SGX-HW then image will be overwritten.

Ram-srini: @manju956 @manojgop I have changed docker image name as suggested, review it.

PR #460, opened 2020-05-12
Add IAS verification to generic client

Ram-srini:
1. Chnages to generic client
2. Fix to verify report python module

### DIFF
--- a/common/crypto_utils/setup.py
+++ b/common/crypto_utils/setup.py
@@ -63,25 +63,29 @@ crypto_include_dirs = [
 
 verify_report_include_dirs = [
     os.path.join(tcf_root_dir, 'common/cpp'),
+    os.path.join(tcf_root_dir, 'common/cpp/crypto'),
+    os.path.join(tcf_root_dir, 'common/cpp/packages/parson'),
+    os.path.join(tcf_root_dir, 'common/cpp/packages/base64'),
     os.path.join(tcf_root_dir, 'common/cpp/verify_ias_report'),
-]
+] + openssl_include_dirs
 
 library_dirs = [
     os.path.join(tcf_root_dir, "common/cpp/build"),
 ] + openssl_lib_dirs
 
+# Do not change the order of these
+# libraries otherwise we will get the undefined
+# symbols in verify_report shared library
 libraries = [
+    'uavalon-verify-ias-report',
+    'uavalon-crypto',
     'uavalon-common',
     'uavalon-base64',
     'uavalon-parson',
-    'uavalon-crypto',
-    'uavalon-verify-ias-report'
 ] + openssl_libs
 
-libraries = libraries + openssl_libs
-
 crypto_modulefiles = [
-    "avalon_crypto_utils/crypto/crypto.i"
+    "avalon_crypto_utils/crypto/crypto.i",
 ]
 
 crypto_module = Extension(
--- a/examples/apps/generic_client/generic_client.py
+++ b/examples/apps/generic_client/generic_client.py
@@ -24,6 +24,8 @@ import secrets
 import config.config as pconfig
 import utility.logger as plogger
 import avalon_crypto_utils.crypto_utility as crypto_utility
+import avalon_crypto_utils.verify_report.verify_attestation_report \
+    as attestation_util
 from avalon_sdk.worker.worker_details import WorkerType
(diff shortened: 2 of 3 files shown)


common/crypto_utils/setup.py
```diff
@@ -63,25 +63,26 @@
...
-]
+] + openssl_include_dirs
 
 library_dirs = [
     os.path.join(tcf_root_dir, "common/cpp/build"),
 ] + openssl_lib_dirs
 
 libraries = [
+    'uavalon-verify-ias-report',
+    'uavalon-crypto',
```
manojgop, line 78: Just curious. Did the order of adding library matter ?

common/crypto_utils/setup.py
```diff
@@ -63,25 +63,26 @@
...
     'uavalon-parson',
-    'uavalon-crypto',
-    'uavalon-verify-ias-report'
 ] + openssl_libs
 
-libraries = libraries + openssl_libs
-
 crypto_modulefiles = [
-    "avalon_crypto_utils/crypto/crypto.i"
+    "avalon_crypto_utils/crypto/crypto.i",
```
manojgop, line 85: Why we need a comma at the end ?

examples/apps/generic_client/generic_client.py
```diff
@@ -178,6 +180,31 @@ def _lookup_first_worker(worker_registry, jrpc_req_id):
...
+        enclave_info = {
+            'verifying_key': worker_obj.verification_key,
+            'encryption_key': worker_obj.encryption_key,
+            'proof_data': worker_obj.proof_data,
+            'enclave_persistent_id': ''
+        }
+
+        logger.info("Perform verification of attestation report")
+        verify_report_status = attestation_util.verify_attestation_report(
+            enclave_info)
```
manojgop, line 200: Client needs to check if MRENCLAVE value of worker matches with proof data before submitting work orders. For that client needs to read MRENCLAVE value of worker from config file.

examples/apps/generic_client/generic_client.py
```diff
@@ -178,6 +180,31 @@ def _lookup_first_worker(worker_registry, jrpc_req_id):
     return worker_id
 
 
+def _do_ias_verification(worker_obj):
+    # Do ias verfication on proof data if it exists
```
danintel, line 184: s/ias/IAS/

common/crypto_utils/setup.py
Ram-srini: yes changing order fixed the undefined symbol error in libuavalon-verify-ias-report.

examples/apps/generic_client/generic_client.py
Ram-srini: Should client has to do it explicitly? Can't this thing done in verify_attestation_report API? Which config has the data about MRENCLAVE?

common/crypto_utils/setup.py
Ram-srini: No. Doesn't make difference it is last element the array.

examples/apps/generic_client/generic_client.py
EugeneYYY: verify_attestation_report API is probably a better place for verifying MRENCLAVE value since it can be more broadly reused.
We also should  consider verifying a) WPE MRENCLAVE (2nd value) b) MRSIGNER.
It makes sense to merge this PR and then create a new PR that will
- modify verify_attestation_report() to accept MRENCLAVE (singleton or KME), MRENCLAVE of 
  WPE, MRSIGNER. If any of the values is not provided (null or empty string), the function 
  should not verify it
- modify _do_IAS_verification() function to pass the new additional parameters. In case of the 
  generic client these parameters can be accepted on the command line or in a config file 
  (the config file name is provided as a command line parameter)

examples/apps/generic_client/generic_client.py
```diff
@@ -178,6 +180,33 @@ def _lookup_first_worker(worker_registry, jrpc_req_id):
     return worker_id
 
 
+def _do_IAS_verification(worker_obj):
```
manojgop, line 183: Can we rename to more generic as `_do_worker_verfication()`. Avalon clients is agnostic about the details of how attestation is happening at back end (IAS or DCAP).  Only Avalon core framework will have the visibility of details of attestation.

PR #461, opened 2020-05-12
k8s setup for Avalon Fabric proxy model

pankajgoyal2:
This PR demostrates
- How to setup k8s cluster for Avalon Fabric proxy model


docker/k8s/fabric/README.md
```diff
@@ -0,0 +1,59 @@
...
+## Running 
+
+1. Start Shared KV database service.
+    ```bash
+    # bootstrap the deployment and service
+    kubectl create -f ../lmdb.yaml
+    ```
+2. Start `EnclaveManager` (to manage Intel SGX enclaves)  
+    ```bash
+    kubectl create -f ../enclave-manager-deployment.yaml
```
manojgop, line 24: This deployment file is for SGX SIM. For SGX HW we need `kubectl create -f ../sgx-enclave-manager-deployment.yaml`. Can we add this step for Intel SGX hardware mode.

docker/k8s/fabric/blockchain-connector.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - name: blockmark
+          mountPath: /project/avalon/blockchain_connector/blockmark
+        - name: keyfiles
+          mountPath: /keyfiles
+        - name: network
+          mountPath: /project/avalon/sdk/avalon_sdk/connector/blockchains/fabric/network.json
+      volumes:
+      - name: blockmark
+        hostPath:
+          path: /home/ubuntu/mywork/vars/blockmark
```
manojgop, line 32: Will this solution work, if block chain connector pods gets deployed in different nodes ? Is this path configured by `start_fabric.sh` ? If so, please add a section in documentation with assumptions and current limitation

docker/k8s/fabric/README.md
pankajgoyal2: For fabric proxy, I referred to docker-compose-fabric.yaml which is only for SGX SIM. If fabric proxy is running well with SGX HW then yes, what you are asking for, it can be done.

docker/k8s/fabric/blockchain-connector.yaml
pankajgoyal2: This is the limitation with both shell and blockchain connector containers. So is with k8s pods.

pankajgoyal2: Though start_fabric.sh allows to install minifab at a place of our choice. I am going to work on that in my next PR to improve on it.

docker/k8s/fabric/README.md
manojgop: Technically it should work as enclave manager is independent of direct/proxy model. But we should test this without fail.

docker/k8s/fabric/blockchain-connector.yaml
manojgop: Initial PR can assume single node setup and we can document this limitation. But we need to discuss how to address this limitation in Proxy model since we need to support k8s pods with multiple nodes. Direct model k8s pods works with multiple nodes.

pankajgoyal2: I agree. This PR doesn't target that. I will be on this task to remove this limitation as a next step.

docker/k8s/fabric/README.md
```diff
@@ -0,0 +1,61 @@
+# Kubernetes playground for Avalon Fabric proxy model
+
+This serves as a playground to setup k8s cluster for Avalon Fabric proxy model.
+
+## Prerequisites
+- To setup the k8s cluster using minikube, refer to Prerequisites section in README.md under docker/k8s folder
```
danintel, line 6: Add a link to README.md in the parent directory:

```
To setup the k8s cluster using minikube, refer to Prerequisites section in
 [README.md](../README.md) under docker/k8s folder
```

docker/k8s/fabric/README.md
```diff
@@ -0,0 +1,61 @@
...
+- This setup is going to work with the assumption that fabric network with minifab, Avalon blockchain connector pod
```
danintel, line 7: s/fabric/the Fabric/

docker/k8s/fabric/README.md
```diff
@@ -0,0 +1,61 @@
...
+  and Avalon shell pod are running in same machine.
```
danintel, line 8: s/in same/in the same/

docker/Dockerfile
```diff
@@ -241,6 +241,9 @@ COPY --from=common_python_image /project/avalon/common/python/dist/*.whl dist/
 COPY --from=common_crypto_utils_image /project/avalon/common/crypto_utils/dist/*.whl dist/
 COPY --from=avalon_sdk_image /project/avalon/sdk/dist/*.whl dist/
 
+# Copy Scripts
```
Ram-srini, line 244: Add comment why do we need to copy this script file into shell container.

scripts/start_fabric.sh
```diff
@@ -145,7 +145,7 @@ elif [ $START_FABRIC == 1 ]; then
     # If already fabric network up and running skip start again.
     if [[ ! $(docker ps --format '{{.Names}}' |grep "peer*") ]]; then
         # Start the fabric containers peers, orderers, cas
-        minifab up -i $FABRIC_VERSION
+        minifab up -i $FABRIC_VERSION -e true
```
Ram-srini, line 148: Add comment what -e do in the script

docker/k8s/fabric/blockchain-connector.yaml
Ram-srini: All we need to share the crypto materials generated from minifab needs to be shared to avalon-shell and blockchain connector then, we are good run in multiple machines.

docker/k8s/fabric/README.md
pankajgoyal2: Yes, it works.

docker/k8s/fabric/README.md
pankajgoyal2: Done

docker/k8s/fabric/README.md
```diff
@@ -0,0 +1,68 @@
+# Kubernetes playground for Avalon Fabric proxy model
+
+This serves as a playground to setup k8s cluster for Avalon Fabric proxy model.
+
+## Prerequisites
+- To setup the k8s cluster using minikube, refer to Prerequisites section in [README.md](../README.md) under docker/k8s folder
+- This setup is going to work with the assumption that Fabric network with minifab, Avalon blockchain connector pod
```
danintel, line 7: Add "the" before "Fabric"

docker/k8s/sgx-enclave-manager-deployment.yaml
```diff
@@ -15,21 +15,21 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: avalon-sgx-enclave-manager
+  name: avalon-enclave-manager
```
manojgop, line 18: In K8s both SGX SIM and SGX HW pods can be running at same time. At least I use it usually for developmnet/testing purpose. Else every time we need to delete the deployment of SGX SIM pods before starting SGX HW pods and vice-versa. Are you facing any issue if we name SGX SIM and SGX HW pods for enclave manager differently ?

pankajgoyal2: I couldn't get your point. When you say SGX SIM and SGX HW pods, are you referring to enclave manager pod?

docker/k8s/README.md
```diff
@@ -54,7 +54,7 @@ in SGX-SIM and SGX-HW mode.
     ```
 
     > The above command would run the 
-    > [propose_requests.sh](../../scripts/propose_requests.sh) as a Kubernetes job
+    > [propose_requests.sh](../../tools/propose_requests.sh) as a Kubernetes job
```
manojgop, line 57: May be it makes more sense to keep the `propose_requests.sh` script in K8s directory.  Please modify the https://github.com/hyperledger/avalon/blob/master/docker/k8s/propose-requests-job.yaml#L17 based on the location of this script

docker/k8s/sgx-enclave-manager-deployment.yaml
manojgop: Yes. But this PR anyways has a service end point for enclave manager. So looks ok. Let's keep it as per this PR for now. If we face any issues later we can change.

tools/propose_requests.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
```
manojgop, line 1: This file was moved to docker/k8s in this PR ?

pankajgoyal2: yes

PR #462, opened 2020-05-12
Use System Call Instead of OpenSSL for Random Number Generation

danintel:
Replaces RAND_bytes() with Linux getrandom(GRND_NONBLOCK) system call.
This reduces dependence on OpenSSL.

### DIFF
--- a/common/cpp/crypto/crypto_utils.cpp
+++ b/common/cpp/crypto/crypto_utils.cpp
@@ -13,16 +13,8 @@
  * limitations under the License.
  */
 
-#include <string.h>
-#include <assert.h>
-#include <openssl/err.h>
-#include <openssl/rand.h>
+#include <sys/random.h>
 #include <openssl/sha.h>
-#include <openssl/x509v3.h>
-#include <openssl/pem.h>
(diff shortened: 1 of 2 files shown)


common/cpp/crypto/crypto_utils.cpp
```diff
@@ -52,27 +44,28 @@ namespace Error = tcf::error;
...
-
-    if (res != 1) {
-        std::string msg("Crypto Error (RandomBitString): ");
-        ERR_load_crypto_strings();
-        ERR_error_string(ERR_get_error(), err);
-        msg += err;
-        throw Error::RuntimeError(msg);
+    // Get random bytes until buffer is full
+    for (ssize_t i = length; i > 0; i =- res) {
+        res = getrandom(&buf.data()[length - i], i, GRND_NONBLOCK);
```
manojgop, line 63: Avalon crypto library is used by Enclave. So systems calls cannot be used. 
https://github.com/hyperledger/avalon/blob/6039768e7314ed1eea5472923559daa2799f061e/common/cpp/crypto/CMakeLists.txt#L52

PR #464, opened 2020-05-12
Add MbedTLS Random and Digest Crypto Code as an Alternative to OpenSSL

danintel:
This first phase adds code for random numbers and SHA-256 digests.

MBedTLS is another cryptographic implementation similar to OpenSSL
in functionality, but with a smaller TCB.
https://www.mbed.com/en/technologies/security/mbed-tls/

This will allow the possibility of using mbedtls-sgx as an alternative to
OpenSSL.

### DIFF
--- a/common/cpp/tests/Makefile
+++ b/common/cpp/tests/Makefile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-CPPFLAGS= -D_UNTRUSTED_ -DCRYPTOLIB_OPENSSL
+CPPFLAGS= -D_UNTRUSTED_
 CPPFLAGS+= -I..  -I../crypto -I../packages/base64
 LDFLAGS=-lcrypto
 
@@ -22,11 +22,15 @@ ifdef CRYPTOLIB_MBEDTLS
 	CPPFLAGS+= -DCRYPTOLIB_MBEDTLS
 	LDFLAGS+=-lmbedcrypto
 endif
+ifdef CRYPTOLIB_OPENSSL
+	CPPFLAGS+= -DCRYPTOLIB_OPENSSL
+endif
 
 UTILTESTOBJS= build/crypto_utils.o build/crypto_utils_encrypt.o \
 		build/skenc.o build/types.o \
 		build/hex_string.o build/utils.o build/base64.o
 
+# First matching pattern build rule found is used
 build/%: build/%.o
 	g++ -o $@ $^ $(LDFLAGS)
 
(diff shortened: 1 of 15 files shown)


danintel:
Unit test results:

```
tests (mbedtls)  $ USE_MBED_TLS=1 make
g++ -o build/utiltest.o -D_UNTRUSTED_ -I..  -I../crypto -I../packages/base64 -c utiltest.cpp
g++ -o build/utiltest build/utiltest.o build/crypto_utils.o build/skenc.o build/types.o build/hex_string.o build/utils.o build/base64.o -lcrypto

tests (mbedtls) $ build/utiltest
Random number test: RandomBitString()
RandomBitString generation PASSED:
80190929719F2077648E351B5969A191B11B87A0FA64F4BDCAA49D08B689E254
RandomBitString PASSED: invalid length detected.
Hash SHA-256 test: ComputeMessageHash()
PASSED: ComputeMessageHash("Hyperledger Avalon"): 
Hash SHA-256 test: ComputeMessageHash()
PASSED: ComputeMessageHash(""): 
Test key generation: CreateHexEncodedEncryptionKey()
Key is: 05B9EAAF151919F6A424917D3EFE5A42F5809CC5C0DC2E84F4E7F2953728A8EB
PASSED: CreateHexEncodedEncryptionKey()
Test encryption: EncryptData()/DecryptData()
PASSED: EncryptData()/DecryptData()
Crypto utility tests PASSED.
```

common/cpp/crypto/crypto_utils.cpp
```diff
@@ -52,28 +51,60 @@ namespace Error = tcf::error;
...
+    int res;
 
     if (length < 1) {
         std::string msg("Crypto Error (RandomBitString): "
             "length argument must be at least 1");
         throw Error::ValueError(msg);
     }
 
-    res = RAND_bytes(buf.data(), length);
+#ifdef USE_MBEDTLS
```
manojgop, line 68: If we are planning to add more mbedtls crypto functions in this file then I would suggest to create a different file for easier maintenance.  Unfortunately Avalon C++ Crypto implementation is not really object oriented else we should could use some design pattern like factory pattern to create OpenSSL or MbedTLS  crypto utils class

danintel: OK, I will have separate files for OpenSSL/Mbed TLS.
I will first file another PR to separate the OpenSSL-dependent and independent parts of the .cpp implementation.

danintel: @manojgop I updated the PR. I separated the Mbed TLS code from OpenSSL code.
More Mbed TLS code is coming--this is the first step.
Please re-review.

common/cpp/tests/Makefile
```diff
@@ -36,10 +37,21 @@ build/%.o: %.cpp
...
 	g++ -o $@ $(CPPFLAGS) -c $^
 
-build/%.o: ../packages/base64/%.cpp
+# Library-specific source has precedence over generic source in ../crypto/
+ifdef CRYPTOLIB_OPENSSL
+build/%.o: ../crypto/openssl/%.cpp
+	g++ -o $@ $(CPPFLAGS) -c $^
+endif
+
+ifdef CRYPTOLIB_MBEDTLS
```
manojgop, line 49: Is it possible to define both CRYPTOLIB_OPENSSL and CRYPTOLIB_MBEDTLS together or either one of them ?

danintel: Yes. I will add a check that both are not defined at the same time.

manojgop: If we define both do we get any errors like multiple symbols defined for functions "RandomBitString" ?

danintel: @manojgop I added a check for not defining both in `crypto_shared.h`.
There are no multiple symbols because only one or the other file is compiled with `test/Makefile` (not both).

PR #467, opened 2020-05-13
Separate OpenSSL-dependent and -independent Parts of C++ Crypto Code

danintel:
- crypto_utils_encrypt.cpp:
  split OpenSSL-independent part from crypto_utils.cpp
- *.cpp: Remove unused header files
- *.cpp: Add Doxygen comments
- *.h:   Remove use of OpenSSL-specific constants in header files
- *.h:   Document private (implementation internal) vs. public header files


EugeneYYY:
1)	All OpenSSL (and in the future MbedTLS) specific files should be conditionally compiled. It will allow to select a crypto lib (e.g. via an environmental variable) at compile time without changes to the makes files (that will list files for both OpenSSl and MbedTLS). Obviously, it won’t prevent using make files fine tuned to use a specific crypto lib. Also see below 2) on how to specify the crypto lib without breaking current make files. All crypto lib specific /h and .cpp files should be wrapped like below (for openssl)
#if CRYPTO_LIB == CRYPTO_LIB_OPENSSL
   ... file content goes here.
#endif

2)	File crypto.h should include a macro defining which crypto library to use. There should be a compilation time check ensuring the library is defined and set to a valid value like below
#define CRYPTO_LIB_OPENSSL  1 
#define CRYPTO_LIB_MBEDTLS  2

#define CRYPTO_LIB CRYPTO_LIB_MBEDTLS

#ifndef CRYPTO_LIB
   #define CRYPTO_LIB CRYPTO_LIB_MBEDTLS
#endif`

#if (CRYPTO_LIB != CRYPTO_LIB_OPENSSL) && (CRYPTO_LIB != CRYPTO_LIB_MBEDTLS)
    #error "invalid CRYPTO_LIB value"
#endif

3)	There should be a crypto lib specific file, e.g. below is an example for crypto_openssl.h (only one macro is to be defined at this time)
#if CRYPTO_LIB == CRYPTO_LIB_OPENSSL
   #pragma once
namespace tcf {
namespace crypto_openssl_constants {
         // Use RSA-3072 for long-term security,
         // with OAEP padding and SHA-256 digest.
         // The 2048 byte buffer allows for future key sizes <= RSA-16384.
         // RSA is not quantum resistant.
        const int RSA_KEY_SIZE = 2048;
        const int RSA_PADDING_SIZE = 41;

        constexpr int RSA_PLAINTEXT_LEN =
            ((RSA_KEY_SIZE - RSA_PADDING_SIZE) >> 3);
}  // namespace crypto_openssl_constants
}  // namespace tcf
#endif

common/cpp/crypto/sig_private_key.cpp
```diff
@@ -204,7 +188,7 @@ void pcrypto::sig::PrivateKey::Generate() {
         throw Error::RuntimeError(msg);
     }
 
-    EC_GROUP_ptr ec_group(EC_GROUP_new_by_curve_name(constants::CURVE),
+    EC_GROUP_ptr ec_group(EC_GROUP_new_by_curve_name(NID_secp256k1),
```
EugeneYYY, line 191: We may want to use different curves so it is better to have a corresponding definition in an crypto lib generic header file. See my generic PR comment #3.

danintel: @EugeneYYY: I restoerd constants::CURVE and moved it to an internal header file, crypto_shared.h so it will not be exposed to non-crypto Avalon code.

danintel: @EugeneYYY: I added CRYPTOLIB_OPENSSL and CRYPTOLIB_MBEDTLS to private header file `crypto_shared.h` and `tests/Makefile`.

I did not add `CRYPTOLIB*` to crypto.h because it is used by crypto code in other parts of Avalon and the API should be library-independent.

I updated the PR, so please re-review.

common/cpp/tests/Makefile
```diff
@@ -12,10 +12,19 @@
...
+CPPFLAGS= -D_UNTRUSTED_ -DCRYPTOLIB_OPENSSL
 CPPFLAGS+= -I..  -I../crypto -I../packages/base64
 LDFLAGS=-lcrypto
-UTILTESTOBJS= build/crypto_utils.o build/skenc.o build/types.o \
+
+# To use Mbed TLS instead of the default OpenSSL, define CRYPTOLIB_MBEDTLS=1
+# For example: CRYPTOLIB_MBEDTLS=1 make
+ifdef CRYPTOLIB_MBEDTLS
+	CPPFLAGS+= -DCRYPTOLIB_MBEDTLS
+	LDFLAGS+=-lmbedcrypto
```
manojgop, line 23: Don't we need conditional compilation for `CPPFLAGS` for OPENSSL as well. `CPPFLAGS= -D_UNTRUSTED_ -DCRYPTOLIB_OPENSSL` is always added. So if `CRYPTOLIB_MBEDTLS` is defined we will have both OPENSSL and MBEDTLS

danintel: @manojgop You are correct--we will need the same for OpenSSL as well. In the transition period, only part of the crypto functions are implemented with Mbed TLS--the rest is OpenSSL.  In other words, there are two choices
- Pure OpenSSL
- Mixed Mbed TLS + OpenSSL

In the future it will be:
- Pure Openssl
- Pure Mbed TLS
But we are not there yet.

PR #475, opened 2020-05-18
Add WPE code to generate Nonce

manju956:
- Generates trusted nonce to be used by WPE to get unique id from KME.
- Persist the nonce in EnclaveData class to use it while verifying the signature of unique id.
- Initialize EnclaveData during enclave initialzation

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave/signup_enclave_wpe.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave/signup_enclave_wpe.cpp
@@ -29,6 +29,7 @@
 #include "zero.h"
 #include "jsonvalue.h"
 #include "parson.h"
+#include "hex_string.h"
 
 #include "enclave_data.h"
 #include "enclave_utils.h"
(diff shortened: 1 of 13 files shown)


tc/sgx/trusted_worker_manager/enclave/signup_enclave_wpe.cpp
```diff
@@ -44,6 +44,17 @@ static void CreateSignupReportDataWPE(const uint8_t* ext_data,
     EnclaveData* enclave_data,
     sgx_report_data_t* report_data);
 
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+tcf_err_t ecall_GenerateNonce(uint8_t* out_nonce, size_t in_nonce_size) {
+    tcf_err_t result = TCF_SUCCESS;
+    tcf::error::ThrowIf<tcf::error::ValueError>(in_nonce_size <= 0,
+        "Nonce size should be positive value");
+
+    sgx_status_t ret = sgx_read_rand(out_nonce, in_nonce_size);
```
manojgop, line 53: Can we use openSSL function to generate nonce. That way this function can be SGX independent.

manju956: Ok. I used crypto utility function which uses openssl/sgxssl to generate random nonce.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_wpe.cpp
```diff
@@ -44,6 +44,29 @@ static void CreateSignupReportDataWPE(const uint8_t* ext_data,
     EnclaveData* enclave_data,
     sgx_report_data_t* report_data);
 
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+tcf_err_t ecall_GenerateNonce(uint8_t* out_nonce, size_t in_nonce_size) {
+    tcf_err_t result = TCF_SUCCESS;
+    tcf::error::ThrowIf<tcf::error::ValueError>(in_nonce_size <= 0,
+        "Nonce size should be positive value");
+    try {
+        out_nonce = tcf::crypto::RandomBitString(in_nonce_size).data();
```
manojgop, line 53: WPE enclave has to store the nonce inside WPE enclave and verify the unique ID signature from KME using this nonce.

manju956: Done.

tc/sgx/trusted_worker_manager/enclave/signup_enclave_wpe.cpp
```diff
@@ -44,6 +45,34 @@ static void CreateSignupReportDataWPE(const uint8_t* ext_data,
...
+tcf_err_t ecall_GenerateNonce(uint8_t* out_nonce, size_t in_nonce_size) {
+    tcf_err_t result = TCF_SUCCESS;
+    tcf::error::ThrowIf<tcf::error::ValueError>(in_nonce_size <= 0,
+        "Nonce size should be positive value");
+    try {
+        ByteArray nonce_bytes = tcf::crypto::RandomBitString(in_nonce_size);
+        // Convert nonce to hex string and persist in the EnclaveData
+        EnclaveData* enclaveData = EnclaveData::getInstance();
+        std::string nonce_hex = ByteArrayToHexEncodedString(nonce_bytes);
+        enclaveData->set_nonce(nonce_hex);
```
manojgop, line 58: WPE needs to check if this nonce is present in signature obtained from KME (uniqe ID + nonce). Are we doing that ?

PR #478, opened 2020-05-19
Refactor work order request handler for KME worker

manju956:


tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge_wrapper/work_order_wrap_kme.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "swig_utils.h"
+#include "types.h"
+
+#include "work_order_kme.h"
+
+#include "base.h"
+#include "work_order_wrap_kme.h"
+
+std::string HandleWorkOrderRequestKME(
+    const std::string& serialized_request,
```
Ram-srini, line 29: Add function comment why do need this wrapper function and what it will do.

tc/sgx/trusted_worker_manager/enclave/work_order_enclave_kme.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "error.h"
+#include "tcf_error.h"
+#include "utils.h"
+
+#include "enclave_utils.h"
+#include "base_enclave.h"
+#include "enclave_data.h"
+#include "work_order_processor.h"
+
+ByteArray last_serialized_response_kme;
```
manojgop, line 32: We need to see how to avoid this global variable. This will get overwritten if KME get requests from multiple WPEs (even if WPEs handles same workload type). Currently the work order is executed in serial order. So it might work. Later on we may need to use map of workorder_id as key and serialized_response_kme as value. We can do in a separate PR.

PR #479, opened 2020-05-19
Increase wait time in proxy model test script

Ram-srini:
Fresh instantiation of chaincode will take more time
and tests can't be triggered till then.


danintel:
Builds are randomly failing and it looks like this will fix it, so I will merge this now.

PR #484, opened 2020-05-19
Update network name to disregard directory

rranjan3:
- Update network name to force use besu_local_net disregarding
  the docker-compose file location. By default docker-compose
  prepends the directory name in to the name of all components.
- Revert static IP removal till a resolution is reached. As of
  of now these IP addresses are part of the genesis file for
  the Besu nodes.

### DIFF
--- a/docs/dev-environments/ethereum/besu/docker-compose.yaml
+++ b/docs/dev-environments/ethereum/besu/docker-compose.yaml
@@ -19,6 +19,7 @@ services:
       - './besu/node1:/var/lib/besu'
     networks:
       local_net:
+        ipv4_address: '172.13.0.2'
 
   #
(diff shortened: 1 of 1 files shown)


rranjan3:
> Approved, but I have a question.
> I assume there's some problem when Docker assigns IP addresses and sets up a network for these containers. So you are hard coding. What problem are you fixing?

@danintel the IP addresses are already hardcoded in the genesis file of the Besu nodes which we were not aware of. What we attempted to do is remove reliance on static IP in the docker network of Besu nodes. The goal was to make these IP addresses of the containers dynamic. This would be a better practice overall. But, an attempt to make them dynamic did not work. The network would hang and no transactions would be mined. We then learnt( with help from @jimthematrix ) that this was because of the static IP hardcoded within the Besu network itself(which is required). Hence, his PR attempted to reinstate the original state where the containers would be assigned static IPs.

PR #489, opened 2020-05-20
Refactor work order request handler

manju956:
- Make uniform work order ecall API signatures for all worker types
- Use function pointers to call appropriate ecalls for different
workers. This eliminates redundancy in code.
- Delete work order files for KME and WPE in enclave bridge and bridge
wrapper.


common/cpp/tcf_error.h
```diff
@@ -45,6 +45,16 @@ typedef enum {
     TCF_ERR_INVALID_WORKLOAD = -12
 } tcf_err_t;
 
+/** Avalon worker enclave Type (singleton, KME, WPE) */
+enum EnclaveType {
+    // SINGLETON_ENCLAVE does both Key management and workload processing
```
danintel, line 50: Doxygen requires one of these formats so it will know to use it in documentation:
`/// SINGLETON_ENCLAVE does both Key management and workload processing`
`//! SINGLETON_ENCLAVE does both Key management and workload processing`
`/** SINGLETON_ENCLAVE does both Key management and workload processing */`

Use one of the above here and on lines 52 and 54.

See 
http://www.doxygen.nl/manual/docblocks.html#cppblock

common/cpp/types.h
```diff
@@ -27,16 +27,6 @@
 #include <vector>
 #include <list>
 
-/** Avalon worker enclave Type (singleton, KME, WPE) */
```
danintel, line 30: Why did you move this to tcf_error.h?
File types.h seems more appropriate as these are not error codes.

common/cpp/types.h
```diff
@@ -27,16 +27,6 @@
...
-enum EnclaveType {
-    // SINGLETON_ENCLAVE does both Key management and workload processing
```
Ram-srini, line 32: types.h would be right place to keep enclaveType enum. Any reason you moved to tcf_error.h?

enclave_manager/setup.py
```diff
@@ -34,6 +34,8 @@
 enclave_bridge_path = os.path.join(tcf_root_dir,
         'tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge')
 
+sgx_sdk = os.environ.get('SGX_SDK', '/opt/intel/sgxsdk')
```
Ram-srini, line 37: Add a comment path '/opt/intel/sgxsdk' is default of SGX SDK/.

common/cpp/types.h
manju956: I did move the enclave type enum to tcf_error.h because there are other enums in place. Also, types.h exposes lot of APIs like ByteArrayToString, ByteArrayToStringArray. This was resulting in undefined symbol errors during python module load time. Hence, i moved it to tcf_error.h

common/cpp/types.h
manju956: I did move the enclave type enum to tcf_error.h because there are other enums in place. Also, types.h exposes lot of APIs like ByteArrayToString, ByteArrayToStringArray along with some typedefs. This was resulting in undefined symbol errors during python module load time. Hence, i moved it to tcf_error.h

enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
```diff
@@ -34,8 +34,10 @@ def execute(self):
         encrypted_request = crypto.byte_array_to_base64(serialized_byte_array)
 
         try:
-            encoded_encrypted_response = \
-                enclave.HandleWorkOrderRequest(encrypted_request)
+            # Work order extended data is not used in Singleton,
+            # hence passing empty value
+            encoded_encrypted_response = enclave.HandleWorkOrderRequest(
+                encrypted_request, "", enclave.SINGLETON_ENCLAVE)
```
manojgop, line 40: Is this file specific to singleton work order handling ? I would suggest to add enclave type in SgxWorkOrderRequest constructor.  By default the value can be SINGLETON_ENCLAVE if it's not initialized.

common/cpp/types.h
manojgop: Please create a new file to add this enum.

enclave_manager/avalon_enclave_manager/singleton/singleton_enclave.i
```diff
@@ -83,12 +83,14 @@ namespace std {
...
 #include "signup_info.h"
 #include "signup_info_singleton.h"
+#include "tcf_error.h"
 %}
 
 %{
 #include "tcf_enclave.h"
 %}
 
+%include "tcf_error.h"
```
manojgop, line 93: Not sure why you need to expose the entire tcf_error.h definitions in Python layer. If it's for the enclave types please create a new file to add this enum.

enclave_manager/setup.py
```diff
@@ -56,8 +59,10 @@
     enclave_bridge_wrapper_path,
     enclave_bridge_path,
     os.path.join(tcf_root_dir, 'common/cpp'),
+    os.path.join(tcf_root_dir, 'tc/sgx/trusted_worker_manager/common'),
```
manojgop, line 62: Why is there a dependency on `tc/sgx/trusted_worker_manager/common`

tc/sgx/trusted_worker_manager/enclave/work_order_enclave.cpp
```diff
@@ -33,6 +33,8 @@ ByteArray last_serialized_response;
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 tcf_err_t ecall_HandleWorkOrderRequest(const uint8_t* inSerializedRequest,
     size_t inSerializedRequestSize,
+    const uint8_t* inWorkOrderExtData,
+    size_t inWorkOrderExtDataSize,
```
manojgop, line 37: Is this file specific to singleton ? Then please add a check to ensure these values passed are NULL

tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/work_order.cpp
```diff
@@ -17,15 +17,111 @@
...
 #include "types.h"
 
 #include "enclave.h"
 #include "base.h"
 #include "work_order.h"
 
-tcf_err_t WorkOrderHandlerBase::GetSerializedResponse(
+WorkOrderHandler::WorkOrderHandler(EnclaveType enclave_type) {
+    // Initialize function pointers
+    if (enclave_type == SINGLETON_ENCLAVE) {
```
manojgop, line 29: This is ok for time being. Later when we have multiple enclaves libraries for Singleton, KME and WPE then each enclave will have similar function names. So use Polymorphism and Virtual Functions instead of Function Pointers later

enclave_manager/setup.py
manju956: Because work_order_wrap.cpp depends on work_order.h (enclave bridge file) which uses header files in tc/sgx/trusted_worker_manager/common.

tc/sgx/trusted_worker_manager/enclave/work_order_enclave.cpp
manju956: Ok

enclave_manager/avalon_enclave_manager/singleton/singleton_enclave.i
manju956: Yes, its included to access c++ enum. I will move it to new file and include it in swig.

enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
manju956: Ok

PR #492, opened 2020-05-21
Enabling SGX-HW mode in CI pipeline

pankajgoyal2:
- This PR enables to be able to build Avalon components in SGX-HW
  mode and run tests.

### DIFF
--- a/Jenkinsfile-sgx-hw
+++ b/Jenkinsfile-sgx-hw
@@ -1,0 +1,107 @@
+#!groovy
+
+// Copyright 2020 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// ------------------------------------------------------------------------------
+
+pipeline {
+    agent {
+        node {
+            label 'slave-1'
+            customWorkspace "workspace/${env.BUILD_TAG}"
+        }
+    }
+
+    options {
+        timestamps()
+        buildDiscarder(logRotator(daysToKeepStr: '31'))
+    }
+
+    environment {
+        ISOLATION_ID = sh(returnStdout: true,
+                          script: 'printf $BUILD_TAG | sha256sum | cut -c1-64').trim()
+        COMPOSE_PROJECT_NAME = sh(returnStdout: true,
+                          script: 'printf $BUILD_TAG | sha256sum | cut -c1-64').trim()
+    }
+
+    stages {
+        stage('Check for Signed-Off Commits') {
+            steps {
+                sh '''#!/bin/bash -l
+                    if [ -v CHANGE_URL ] ;
+                    then
+                        temp_url="$(echo $CHANGE_URL |sed s#github.com/#api.github.com/repos/#)/commits"
+                        pull_url="$(echo $temp_url |sed s#pull#pulls#)"
+
+                        IFS=$'\n'
+                        for m in $(curl -s "$pull_url" | grep "message") ; do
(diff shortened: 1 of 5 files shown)


Jenkinsfile-sgx-hw
```diff
@@ -0,0 +1,107 @@
...
+                        unset IFS;
+                    fi
+                '''
+            }
+        }
+
+        stage('Run Lint') {
+            steps {
+                sh 'docker-compose -f docker-compose-lint.yaml up --exit-code-from avalon-lint'
+                sh './../sgx-hw.sh config/singleton_enclave_config.toml'
```
manojgop, line 67: Is `sgx-hw.sh` a new file.  It's missing in this PR

pankajgoyal2: this file is placed in slave machine. It is used to handle usage of specific spid and ias_api_key parameters in enclave config file

manojgop: But still it make sense to include in Git repository since without this SGX HW CI build will not be functional

ci/docker-compose-direct-model-sgx-build.yaml
```diff
@@ -0,0 +1,35 @@
...
+      context: ..
+      dockerfile: docker/Dockerfile
+      args:
+        - http_proxy
+        - https_proxy
+        - no_proxy
+
+  avalon-enclave-manager:
+    image: avalon-sgx-enclave-manager:${ISOLATION_ID}
+    build:
```
manojgop, line 30: Is only 2 containers required. Couldn't understand why this file is named as 
sgx-build as it doesn't seem to build all required Avalon components

Jenkinsfile-sgx-hw
pankajgoyal2: No need. This script is placed in slave machine. Jenkins will trigger the SGX-HW job on the slave where this script is already present. I have tested it and it is working.

ci/docker-compose-direct-model-sgx-build.yaml
pankajgoyal2: If you see Jenkinsfile-sgx-hw in this PR, I am building avalon components like:
docker-compose -f ci/docker-compose-direct-model-build.yaml -f ci/docker-compose-direct-model-sgx-build.yaml build

Jenkinsfile-sgx-hw
manojgop: I mean we don't have a separate repository to maintain the scripts needed by Jenkins slave machines. So we can add all Jenkins server and slave related files in this repo. We can add a README which can mention about slave, server files.

bin/run_tests
```diff
@@ -27,28 +27,35 @@ PATH=$PROJECT_DIR/bin:$PATH
...
-while getopts "dph" opt; do
+while getopts "dpsh" opt; do
     case $opt in
         d)
             DIRECT_MODEL_TESTS=1
             ;;
         p)
             PROXY_MODEL_TESTS=1
             ;;
+	s)
```
Ram-srini, line 51: Why do we need this option? There is no difference in testing with respect to HW/SIM mode.

Jenkinsfile-sgx-hw
```diff
@@ -0,0 +1,107 @@
+#!groovy
+
+// Copyright 2020 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
pankajgoyal2: I am going with this way as first level. An an improvement I will explore it further.

bin/run_tests
pankajgoyal2: there is additional volume map and device map in case of SGX HW mode for avalon-enclave-manager service

Jenkinsfile-sgx-hw
```diff
@@ -0,0 +1,107 @@
...
+            steps {
+                sh 'INSTALL_TYPE="" ./bin/run_tests -s'
+            }
+        }
+
+        stage('Create git archive') {
+            steps {
+                sh '''
+                    REPO=$(git remote show -n origin | grep Fetch | awk -F'[/.]' '{print $6}')
+                    git archive HEAD --format=zip -9 --output=$REPO.zip
```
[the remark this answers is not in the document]
pankajgoyal2: yes, we are using it.

PR #498, opened 2020-05-23
Add container dependency among Besu nodes

rranjan3:
- Make node 2 of besu network to depend on node 1.
  Startup dependency absence causes race condition
  where if the 1st block is produced/published before
  the Synchronizer starts on the other node, the chain
  does not progess.

### DIFF
--- a/docs/dev-environments/ethereum/besu/docker-compose.yaml
+++ b/docs/dev-environments/ethereum/besu/docker-compose.yaml
@@ -58,6 +58,8 @@ services:
       - 23002:8546
     volumes:
       - './besu/node2:/var/lib/besu'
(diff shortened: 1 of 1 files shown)


docs/dev-environments/ethereum/besu/docker-compose.yaml
```diff
@@ -58,6 +58,8 @@ services:
       - 23002:8546
     volumes:
       - './besu/node2:/var/lib/besu'
+    depends_on:
+      - node1.avalon.local
```
Ram-srini, line 62: Add a comment saying node1.avalon.local is created using startup script.

PR #493, opened 2020-05-21
Initial KME/WPE intergation

rranjan3:
- Update configs to include enclave_type
- Enable lint to exclude all setup.*py files
- Minor bug fixes to ensure flow
- Enhancements to wpe requester
- Enhance sgx work order request to include modules dynamically
- Use nonce generated in trusted code in wpe manager
- Minor bug fix in trusted code

PR #500, opened 2020-05-27
Add initial handling at KME listener

rranjan3:
- Add initial handling and creation of work order request
  at KME Listener
- Address remnant review comments from PR #493
- Make kme_listener_url as command line arg as well

### DIFF
--- a/config/wpe_config.toml
+++ b/config/wpe_config.toml
@@ -78,4 +78,4 @@ enclave_type = "WPE"
 [KMEListener]
 # URL of KMEListener where a Work order processor can send requests meant for
 # the Key Management Enclave(KME)  
-bind = "http://avalon-kme:1948"
+kme_listener_url = "http://localhost:1948"
--- a/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
+++ b/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
@@ -28,7 +28,8 @@ from avalon_enclave_manager.base_enclave_manager import EnclaveManager
 from avalon_enclave_manager.worker_kv_delegate import WorkerKVDelegate
 from avalon_enclave_manager.work_order_kv_delegate import WorkOrderKVDelegate
 from listener.base_jrpc_listener import parse_bind_url
-from avalon_enclave_manager.kme.kme_listener import KMEListener
+from avalon_enclave_manager.kme.kme_listener \
+    import KMEListener, construct_wo_req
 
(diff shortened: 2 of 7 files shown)


config/wpe_config.toml
```diff
@@ -78,4 +78,4 @@ enclave_type = "WPE"
 [KMEListener]
 # URL of KMEListener where a Work order processor can send requests meant for
 # the Key Management Enclave(KME)  
-bind = "http://avalon-kme:1948"
+kme_listener_url = "http://avalon-kme:1948"
```
Ram-srini, line 81: Please add command argument to WPE manager as well. If command line argument is passed it takes preference. In docker mode we pass argument and in config file by default it will be localhost.. This way we can handle both stand alone and docker based builds..

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -152,21 +153,44 @@ def _start_kme_listener(self):
...
 
     def RegisterWorkOrderProcessor(self, **params):
         """
         """
+        workload_id = "_kme-reg"
+        attestation_report = params["attestation_report"]
+        in_data = json.dumps({"attestation_report": attestation_report})
+        wo_req = construct_wo_req(in_data, workload_id, self.encryption_key)
+
+        # @TODO : Trusted implementation to be integrated
```
Ram-srini, line 176: FYI registration needs one more argument "verification_key" go it in first step(GetUniqueId)

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
```diff
@@ -83,19 +84,22 @@ def register_wo_processor(self, attestation_report):
...
-    def preprocess_work_order(self, wo_request):
+    def preprocess_work_order(self, wo_request, encryption_key):
         """
         Request to preprocess a work order
 
         Parameters :
-            @param wo_request - The original work order request and
-                                inWorkOrderExData
+            @param wo_request - The original work order request
+            @param encryption_key - WPE's pubblic encryption key
```
danintel, line 93: s/pubblic/public/

config/wpe_config.toml
danintel: I prefer that convention too--localhost as the default and overridden in Docker.

rranjan3: Updated @danintel  @Ram-srini .

PR #503, opened 2020-05-29
Integrate GetVerificationKey at KME

rranjan3:
- Enable call to trusted code for GetVerificationKey
- Fix missing header file in kme enclave swig
- Introduce instance variable in wor_order_processor
  to hold enclave type information
- update kme_workload_plugin to put verification_key
  along with signature in single out-data
- Add response signature verification to KME listener


enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -156,12 +159,29 @@ def GetUniqueVerificationKey(self, **params):
...
+        if "result" not in wo_response_json:
+            logger.error("Could not get UniqueVerificationKey")
+            return wo_response_json
+        if verify_res_signature(wo_response_json['result'],
+                                self.verifying_key):
+            decrypted_res = crypto_utils.decrypted_response(
+                wo_response_json['result'], session_key, session_iv)
+            logger.info("Decrypted response: {}".format(decrypted_res))
+            # Response contains an array of results. In this case, the
+            # array has single element and the data field is of interest.
```
manju956, line 182: Can we document the format of data in the response?

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -265,8 +265,14 @@ namespace tcf {
             // Convert workload_id from hex string to string
             ByteArray workload_bytes = HexStringToBinary(workload_id);
             std::string workload_type(workload_bytes.begin(), workload_bytes.end());
-            WorkloadProcessor *processor = 
-                WorkloadProcessor::CreateWorkloadProcessor(workload_type);
+            WorkloadProcessor *processor;
+            if (this->enclave_type == KME_ENCLAVE){
```
Ram-srini, line 269: Better add comment here why adding this condition.

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -79,25 +79,28 @@ def _manager_on_boot(self):
         self._wo_kv_delegate.cleanup_work_orders()
 
 # -------------------------------------------------------------------------
-
-    def _execute_work_order(self, input_json_str, indent=4):
+    def _execute_work_order(self, input_json_str, ext_data, indent=4):
```
manju956, line 82: can we put default value to ext_data? Not all KME workload operations need it.

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -187,7 +211,7 @@ def PreProcessWorkOrder(self, **params):
         encryption_key = params["encryption_key"]
 
         # @TODO : Trusted implementation to be integrated
-        # wo_response = self._execute_work_order(wo_req, encryption_key)
+        # wo_response = self._execute_work_order(wo_request, encryption_key)
```
danintel, line 214: This is commented out, but it should be wo_req, not wo_request

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -79,25 +79,29 @@ def _manager_on_boot(self):
...
-    def _execute_work_order(self, input_json_str, indent=4):
+    def _execute_work_order(self, input_json_str, ext_data="", indent=4):
         """
         Submits request to KME and retrieves the response
 
         Parameters :
-            input_json_str - A JSON formatted str of the request to execute
+            @param input_json_str - A JSON formatted str of the request
+            @param ext_data - Extended data to pass to trusted code
+            @param indent - Indentation for JSON. Defaults to 4
```
manojgop, line 89: Is indent parameter a configurable in config file. If not, I'm not sure why we need it as a argument to this function as at some place it's hard coded either by caller or have to be passed to functions at all levels to maintain uniformity. I would rather hardcode this when we call json.dumps()

enclave_manager/avalon_enclave_manager/kme/kme_listener.py
```diff
@@ -50,31 +52,31 @@ def __init__(self, rpc_methods):
         super().__init__(rpc_methods)
 
 
-def construct_wo_req(in_data, workload_id, encryption_key):
+def construct_wo_req(in_data, workload_id, encryption_key,
+                     session_key, session_iv):
```
manojgop, line 56: Not sure why `construct_wo_req` defined in  kme listener. AFAIK KME Listener will receive the work order request and WPE will sent the work order request to KME Listener

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -156,12 +160,31 @@ def GetUniqueVerificationKey(self, **params):
...
-
-        # @TODO : Trusted implementation to be integrated
-        # wo_response = self._execute_work_order(json.dumps(wo_req))
 
-        return ""
+        # Create session key and iv to sign work order request
+        session_key = crypto_utils.generate_key()
+        session_iv = crypto_utils.generate_iv()
+
+        wo_req = construct_wo_req(in_data, workload_id, self.encryption_key,
```
manojgop, line 168: No sure what are we trying to do here. My understanding is GetUniqueVerificationKey work order request should be created by WPE sent to KME listener. This work order request creation logic should be in WPE

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
rranjan3: @danintel am I missing something? @212, the variable name is wo_request, the same as this one.

enclave_manager/avalon_enclave_manager/kme/kme_listener.py
rranjan3: The construction of a work order request requires an encryption key. When the WPE requests for UniqueVerificationKey, it is yet to signup and hence does not hold an encryption key. Only after getting the UniqueVerificationKey from the KME does the WPE signup and create the related data.

manojgop: The encryption key used to encrypt inData in wo_request from WPE is the session key generated by WPE and the session key of WPE is encrypted using public encryption key of KME and this encryptedSessionKey is also part of wo_request. WPE needs to looks the worker details of KME to get the KME public encryption key

rranjan3: Plausible. I did think about this. But there could be multiple workers in the KV. What would be the basis of filtering out one worker from the workers available in that case? At  this point the WPE holds only the address of the KMEListener.

manojgop: The expectation is that WPE knows about the KME worker which can handle the workload. In fact this assumption that's true for Avalon clients also i.e Avalpn clients knows which worker can handle teh workload . You could hard code the KME worker id and WPE can retrieve the details of KME worker

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
danintel: @rranjan3  You are correct.  Line 212 was hidden so I did not see it--I saw wo_req at line 199.

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -79,25 +78,28 @@ def _manager_on_boot(self):
...
+            @param input_json_str - A JSON formatted str of the request
+            @param ext_data - Extended data to pass to trusted code
         Returns :
-            json_response - A JSON formatted str of the response received from
-                            the enclave
+            @returns json_response - A JSON formatted str of the response
+                                     received from the enclave
         """
         try:
+            logger.info("Request sent to KME %s", input_json_str)
```
manojgop, line 93: Did you mean Request received by KME ?

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
```diff
@@ -41,73 +46,157 @@ def __init__(self, config):
...
-            return response["result"]
+            wo_response_json = response["result"]
+            if self._verify_res_signature(wo_response_json, verifying_key):
+                decrypted_res = crypto_utils.decrypted_response(
+                    wo_response_json, session_key, session_iv)
+                # Response contains an array of results. In this case, the
+                # array has single element and the data field is of interest.
+                # The data contains result,verification_key and
+                # verification_key_signature delimited by ' '.
+                return decrypted_res[0]['data']
```
manojgop, line 88: Why didn't we use multiple index fields in outData to store verification_key and verification_key_signature ?

enclave_manager/avalon_enclave_manager/wpe/wpe_enclave_manager.py
```diff
@@ -61,15 +60,19 @@ def _create_signup_data(self):
         signup_data = enclave_info.WorkOrderProcessorEnclaveInfo(
             self._config.get("EnclaveModule"))
         self._wpe_requester = WPERequester(self._config)
+        self._worker = self._worker_kv_delegate.get_worker()
```
manojgop, line 63: The worker id should corresponds to the listener end point. WPE can add worker id as a config parameter in wpe_config.toml. We should hard code the worker id string

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
```diff
@@ -41,73 +46,157 @@ def __init__(self, config):
...
+            @returns A json request prepared using parameters passed
+        """
+        # Create work order
+        # Convert workloadId to hex
+        workload_id_hex = workload_id.encode("UTF-8").hex()
+        work_order_id = secrets.token_hex(32)
+        requester_id = secrets.token_hex(32)
+        requester_nonce = secrets.token_hex(16)
+        # worker id is not known here. Hence passing a random string
+        worker_id = secrets.token_hex(32)
```
manojgop, line 180: worker ID should be ID of KME worker corresponding to this WPE. Please read it from config file

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
```diff
@@ -41,73 +46,157 @@ def __init__(self, config):
...
+        # Create session key and iv to sign work order request
+        session_key = crypto_utils.generate_key()
+        session_iv = crypto_utils.generate_iv()
+
+        wo_req = self._construct_wo_req(
+            in_data, workload_id, encryption_key, session_key, session_iv)
+
         json_rpc_request = self._get_request_json("GetUniqueVerificationKey")
-        json_rpc_request["params"] = {"nonce": nonce}
+        json_rpc_request["params"] = {"wo_request": wo_req}
```
manojgop, line 76: Not sure what are we trying to do here. We have to send  `wo_req`  string obtained from `_construct_wo_req()` to KME listener

`response = self._post_and_get_result(wo_req)`

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
```diff
@@ -41,73 +46,157 @@ def __init__(self, config):
...
+            data_encryption_algorithm="AES-GCM-256"
+        )
+        wo_params.add_in_data(in_data)
+
+        # Encrypt work order request hash
+        wo_params.add_encrypted_request_hash()
+
+        return {
+            "jsonrpc": "2.0",
+            "method": workload_id,
```
manojgop, line 196: Pass the method_id here as argument to this function and use it . Like "RegisterWorkOrderProcessor", "PreProcessWorkOrder" etc

rranjan3: Yes. This is how it is being done.

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
rranjan3: The request always does not go alone. There could be extra fields to be sent as ext_data going along. hence, this structure.

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
rranjan3: Added a commit.

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
rranjan3: There is a limitation of sending one outdata per indata field.

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
manojgop: Please send ext_data as inData fields. Please use standard TCF format. Do you see any concerns in using standard TC format to send work order request to KME.

enclave_manager/avalon_enclave_manager/wpe/wpe_requester.py
manojgop: I'm not aware of any such limitation in TCF. inData and outData is interpreted by client and workload

tc/sgx/trusted_worker_manager/enclave/work_order_processor.h
```diff
@@ -31,6 +32,7 @@ namespace tcf {
             ByteArray CreateErrorResponse(int err_code, const char* err_message);
             ByteArray Process(EnclaveData* enclaveData, std::string json_str);
             std::string ext_work_order_data;
+            EnclaveType enclave_type = SINGLETON_ENCLAVE;
```
manojgop, line 35: Can we initialize enclave_type  in constructor or as constructor initializer list

PR #518, opened 2020-06-08
Optimize docker-compose for pool and update document

rranjan3:
- Split docker-compose files for reuse
- Update documentation to update running in pool
  mode

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -48,27 +48,57 @@ Follow the instructions below to execute a Docker-based build and execution.
    `avalon` source repository.
 
    **Intel SGX Simulator mode (for hosts without Intel SGX)**:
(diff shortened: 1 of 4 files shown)


BUILD.md
```diff
@@ -48,27 +48,57 @@ Follow the instructions below to execute a Docker-based build and execution.
    `avalon` source repository.
 
    **Intel SGX Simulator mode (for hosts without Intel SGX)**:
-   1. Run `sudo docker-compose up --build`
+   1. Run:
```
danintel, line 51: Change this to something like:

```
To run in Singleton mode (the same worker handles both keys and workloads):
```

rranjan3: Makes sense.

PR #521, opened 2020-06-09
Add support to send and process input data as plain text

manojgop:
- Enhance generic client to provide an option to send input data
  as plain text.
- Fix issue in enclave work order processor to handle plain text
  in inData. Avoid copying inData keys to outData.
  outData should use client session key by default if client has
  not provided data keys to encrypt outData.


manojgop:
> Add an example to `examples/apps/generic_client/README.md`
> Also add the new option to README.md

Added

examples/apps/generic_client/README.md
```diff
@@ -123,3 +124,10 @@ With Docker:
     --address "0x9Be28B132aeE1b2c5A1C50529a636cEd807842cd" --mode "listing" \
     --workload_id "echo-result" --in_data "Hello"
 ```
+
+### Echo workload with input data as plain text
```
danintel, line 128: I was not sure what "plain text" means. It could mean not in JSON format for example.  Please add something like this:

`### Echo workload with input data as unencrypted plain text`

PR #524, opened 2020-06-10
2 step verification changes

Ram-srini:
If the verification key is present in the work order response, its signature is verified in two steps
1. The response signature is verified using the verification key in the response
2. The verification key signature from the response is verified using worker’s public verification key
(aka KME’s verification key)

### DIFF
--- a/common/crypto_utils/avalon_crypto_utils/signature.py
+++ b/common/crypto_utils/avalon_crypto_utils/signature.py
@@ -322,36 +322,33 @@ class ClientSignature(object):
 
         return input_json_str, SignatureStatus.PASSED
 
-# -----------------------------------------------------------------------------
-    def verify_signature(self, input_json, verification_key):
+    def _verify_wo_response_signature(self, wo_response,
+                                      wo_res_verification_key):
         """
-        Function to verify the signature received from the enclave
+        Function to verify the work order response signature
         Parameters:
-            - input_json is dictionary contains payload returned by the
-              Worker Service in response to successful workorder submit request
-              as per Trusted Compute EEA API 6.1.2 Work Order Result Payload
-            - verification_key is ECDSA/SECP256K1 public key used to verify
-              signatures created by the Enclave.
+            @param wo_response - dictionary contains work order response
+            as per Trusted Compute EEA API 6.1.2 Work Order Result Payload
+            @param wo_res_verification_key - ECDSA/SECP256K1 public key
+            used to verify work order response signature.
         Returns enum type SignatureStatus
         """
-
-        input_json_params = input_json
-
-        nonce = (input_json_params['workerNonce']).encode('UTF-8')
-        signature = input_json_params['workerSignature']
-
+        worker_nonce = (wo_response["workerNonce"]).encode('UTF-8')
+        signature = wo_response['workerSignature']
         hash_string_1 = self.__calculate_hash_on_concatenated_string(
-            input_json_params, nonce)
-        data_objects = input_json_params['outData']
+            wo_response, worker_nonce)
+        data_objects = wo_response['outData']
         hash_string_2 = self.calculate_datahash(data_objects)
         concat_string = hash_string_1 + hash_string_2
         concat_hash = bytes(concat_string, 'UTF-8')
         final_hash = crypto.compute_message_hash(concat_hash)
 
         try:
-            _verifying_key = crypto.SIG_PublicKey(verification_key)
+            _verifying_key = crypto.SIG_PublicKey(wo_res_verification_key)
         except Exception as error:
-            logger.info("Error in verification key : %s", error)
+            logger.error("Error in verification key of "
+                         "work order response : %s",
+                         error)
             return SignatureStatus.INVALID_VERIFICATION_KEY
 
         decoded_signature = crypto.base64_to_byte_array(signature)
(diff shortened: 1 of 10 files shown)


docker-compose-pool.yaml
```diff
@@ -63,8 +63,11 @@ services:
     expose:
       # ZMQ socket port.
       - 5555
+      # sleep is required to wait WPE until KME starts and register worker.
+      # This is temporary work around.
     command: |
       bash -c "
+        sleep 5s
```
manojgop, line 70: This should be handled by WPE Enclave Manager

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -323,35 +323,69 @@ def generate_client_signature(
...
 # -----------------------------------------------------------------------------
-    def verify_signature(self, input_json, verification_key):
+    def verify_signature(self, wo_response, verification_key,
+                         requester_nonce):
         """
         Function to verify the signature received from the enclave
         Parameters:
-            - input_json is dictionary contains payload returned by the
+            @param wo_response - dictionary contains payload returned by the
               Worker Service in response to successful workorder submit request
```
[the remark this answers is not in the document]
Ram-srini: worker nothing but WPE here. KME is Key manager not a worker.

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -323,35 +323,69 @@ def generate_client_signature(
...
+        # step1 - The verification key signature from the
+        # response is verified using worker’s public verification key
+        # (aka KME's verification key)
+        if "extVerificationKey" in wo_response:
+            decrypted_v_key = crypto.base64_to_byte_array(
+                wo_response["extVerificationKey"])
+            v_key = crypto.byte_array_to_string(decrypted_v_key)
+            if requester_nonce is None:
+                logger.info("Missing requester_nonce argument")
+                SignatureStatus.FAILED
```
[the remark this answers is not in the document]
Ram-srini: Yes it is return.

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -323,35 +323,69 @@ def generate_client_signature(
...
+                return SignatureStatus.INVALID_VERIFICATION_KEY
+            decoded_v_key_sig = crypto.base64_to_byte_array(v_key_sig)
+            sig_result = _verifying_key.VerifySignature(
+                v_key_hash,
+                decoded_v_key_sig)
+            if sig_result != 1:
+                logger.info("Error in verifying "
+                            "verification key signature")
+                return SignatureStatus.FAILED
+        else:
```
manojgop, line 371: Can we move else part of another function. This will help reduce complexity of this function. Can we make requester_nonce optional. That way existing applications need not change. requestor_nonce should be present only for extVerificationKeySignature. something like below

```
def verify_signature(self, wo_response, verification_key,requester_nonce=None):

  if "extVerificationKey" in wo_response:
      self.verify_worker_signature_ext(wo_response, verification_key,requester_nonce)
  else:
      self.verify_worker_signature(wo_response,verification_key)
```

Ram-srini: What if worker pool missed that argument? One function handling both the cases(singleton and worker pool) One API

manojgop: If worker pool doesn't pass `requestor_nonce` argument then return error in `verify_worker_signature_ext()`. For singleton case requestor_nonce is not required. We can still keep one function to handle both cases

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -323,35 +323,69 @@ def generate_client_signature(
         return input_json_str, SignatureStatus.PASSED
 
 # -----------------------------------------------------------------------------
-    def verify_signature(self, input_json, verification_key):
+    def verify_signature(self, wo_response, verification_key,
```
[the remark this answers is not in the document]
Ram-srini: Both keys are generated by KME, better name would be wo_response_verification_key

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -323,35 +323,69 @@ def generate_client_signature(
...
+        v_key = None
+        # if verification_key present in work order response
+        # then do 2 step verification
+        # step1 - The verification key signature from the
+        # response is verified using worker’s public verification key
+        # (aka KME's verification key)
+        if "extVerificationKey" in wo_response:
+            decrypted_v_key = crypto.base64_to_byte_array(
+                wo_response["extVerificationKey"])
+            v_key = crypto.byte_array_to_string(decrypted_v_key)
```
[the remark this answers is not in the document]
manju956: extVerificationKeySignature is computed on the hash of the concatenation of b64 encoded verification key and requester nonce hex string.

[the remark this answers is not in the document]
manojgop: pk. Then we should use extVerificationKey instead of vkey. 
Is extVerificationKey  in base64 encoded format.?
If so,  concat_string = extVerificationKey  + requester_nonce ?

docker-compose-pool.yaml
Ram-srini: yes

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -365,6 +362,91 @@ def verify_signature(self, input_json, verification_key):
...
+            status = self._verify_wo_verification_key_signature(
+                wo_response,
+                wo_res_verification_key,
+                requester_nonce)
+            if status == SignatureStatus.PASSED:
+                # step2 : work order response signature is verified
+                # using the verification key in the response
+                decoded_v_key = crypto.base64_to_byte_array(
+                    wo_response["extVerificationKey"])
+                v_key = crypto.byte_array_to_string(decoded_v_key)
```
manojgop, line 436: We can use `extVerificationKey` directly since it's a base64 encoded string.  We don't need `decoded_v_key ` and `v_key` since there is no need to convert 
base64 string -> bytes -> string

Pass `extVerificationKey`  from the response as argument to `_verify_wo_response_signature()`

common/crypto_utils/avalon_crypto_utils/signature.py
```diff
@@ -365,6 +362,91 @@ def verify_signature(self, input_json, verification_key):
...
+            order request. Required in 2 step verification.
+        Returns enum type SignatureStatus
+        """
+        if requester_nonce is None:
+            logger.error("Missing requester_nonce argument")
+            return SignatureStatus.FAILED
+
+        decoded_v_key = crypto.base64_to_byte_array(
+                wo_response["extVerificationKey"])
+        v_key = crypto.byte_array_to_string(decoded_v_key)
```
manojgop, line 385: We can use `extVerificationKey` directly since it's a base64 encoded string. We don't need decoded_v_key and v_key since there is no need to convert
base64 string -> bytes -> string

common/crypto_utils/avalon_crypto_utils/signature.py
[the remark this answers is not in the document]
Ram-srini: Hash is calculated on extVerificationKey + requester_nonce

common/crypto_utils/avalon_crypto_utils/signature.py
Ram-srini: It was needed because verification_key is passed in base64 format..

common/crypto_utils/avalon_crypto_utils/signature.py
Ram-srini: It was needed because verification_key is passed in base64 format. We need plain key with begin and end to verify signature.

common/crypto_utils/avalon_crypto_utils/signature.py
manojgop: @manju956 could you please confirm which format is used by WPE. AFAIK we need to use public key in base64 string format and append that with requester nonce. https://github.com/hyperledger/avalon/blob/master/common/sgx_workload/workload/ext_work_order_info_kme.cpp#L256

PR #529, opened 2020-06-12
Fix verification key encoding format

manju956:
- Change verification_key type from ByteArray to Base64EncodedString(PEM format). This avoids 
   unnecessary data format conversions at multiple places.


common/sgx_workload/workload/ext_work_order_info_kme.cpp
```diff
@@ -249,8 +249,7 @@ int ExtWorkOrderInfoKME::CreateWorkOrderKeyInfo(
         public_sig_key = private_sig_key.GetPublicKey();
 
         ByteArray signing_key = StrToByteArray(private_sig_key.Serialize());
-        wo_key_info.verification_key = StrToByteArray(
-            public_sig_key.Serialize());
+        wo_key_info.verification_key = public_sig_key.Serialize();
```
manojgop, line 252: Please add a comment this is pem formatted string

manju956: Done

PR #530, opened 2020-06-12
Fix stand alone mode test failure

Ram-srini:
Since listener config is hard-coded with container service name
and it is failing in standlone mode.
Added command line option to listener to read zmq_url and default value
in config file is localhost and command line argument takes precedence.

### DIFF
--- a/listener/avalon_listener/tcs_listener.py
+++ b/listener/avalon_listener/tcs_listener.py
@@ -27,6 +27,7 @@
 import sys
 import logging
 import argparse
+from urllib.parse import urlparse
 
 from avalon_listener.tcs_work_order_handler import TCSWorkOrderHandler
 from avalon_listener.tcs_work_order_handler_sync import TCSWorkOrderHandlerSync
@@ -72,8 +73,7 @@ class TCSListener(BaseJRPCListener):
             self.workorder_handler = TCSWorkOrderHandlerSync(
                 self.kv_helper,
                 config["Listener"]["max_work_order_count"],
-                config["Listener"]["zmq_url"],
-                config["Listener"]["zmq_port"])
+                config["Listener"]["zmq_url"])
         else:
             self.workorder_handler = TCSWorkOrderHandler(
                 self.kv_helper,
@@ -121,6 +121,12 @@ def parse_command_line(config, args):
         '--bind', help='URI to listen for requests ', type=str)
     parser.add_argument(
         '--lmdb_url', help='DB url to connect to LMDB ', type=str)
+    parser.add_argument(
+        '--sync_mode', help='Work order execution in synchronous mode',
+        type=bool, default=False)
+    parser.add_argument(
+        '--zmq_url',
+        help='ZMQ url to connect to enclave manager ', type=str)
 
     options = parser.parse_args(args)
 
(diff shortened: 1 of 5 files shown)


listener/avalon_listener/tcs_listener.py
```diff
@@ -150,6 +157,21 @@ def parse_command_line(config, args):
             logger.error("Quit : remote_storage_url is not \
                             present in config for Listener")
             sys.exit(-1)
+    if options.zmq_url:
+        if not is_sync:
+            logger.error("Invalid option zmq_url! It should be supported"
+                         "in work order sync mode")
```
danintel, line 163: The wording is confusing. I would say:

```
logger.error("Invalid option zmq_url! It is only supported"
                         "in work order sync mode")
```

listener/avalon_listener/tcs_listener.py
```diff
@@ -150,6 +157,21 @@ def parse_command_line(config, args):
...
+            sys.exit(-1)
+        else:
+            if config.get("Listener") is None or \
+                        config["Listener"].get("zmq_url") is None:
+                logger.error("Quit : no zmq_url config found for Listener")
+                sys.exit(-1)
+            parse_res = urlparse(options.zmq_url)
+            if parse_res.scheme != "tcp" or parse_res.port == "":
+                logger.error("Invalid zmq url. It should tcp://<host>:<port>")
```
danintel, line 172: s/It should/It should be/

listener/avalon_listener/tcs_listener.py
Ram-srini: zmq_url option is supported only in work order sync execution mode. In asynchronous/pull mode this option doesn't make sense and better to throw error.

pankajgoyal2: is_sync should be the deciding factor for sync mode ON/OFF. having zmq_url is mandatory to pass if is_sync is true otherwise zmq_url should not have any effect. This way you don't have to modify docker-compose files to modify avalon-listener service to remove --zmq_url if you wish to make is_sync false.

Ram-srini: currently work order sync mode is set in toml file and it is default option is on. So either case we have to handle this error. Since default is ON so we have to pass this argument.

pankajgoyal2: I understand that. But is someone wants to disable the sync, then what will be the procedure to get the docker/standalone builds?

ci/avalon-direct-model-sgx-tests.yaml
```diff
@@ -71,9 +71,10 @@ services:
       - 5555
     command: |
       bash -c "
-        avalon_listener --bind http://avalon-listener:1947 --lmdb_url http://avalon-lmdb:9090
+        avalon_listener --bind http://avalon-listener:1947 --lmdb_url http://avalon-lmdb:9090 --zmq_url tcp://avalon-enclave-manager:5555
```
manojgop, line 74: In docker mode, can we read the zmq url from config file . Default zmq url can be tcp://avalon-enclave-manager:5555.   Only in standalone mode we can pass the zmq_url as command line arguent. For worker pool, we need to use polling mode. So passing zmq url in docker mode is not required.

Ram-srini: Either is fine. I am following consistent approach where other services are following like listener and lmdb.

listener/avalon_listener/tcs_listener.py
Ram-srini: Whenever we disable to sync mode then we have to remove those additional parameter, that doesn't make sense.

ci/avalon-direct-model-sgx-tests.yaml
manojgop: Zmq url is needed only in sync mode and in case of docker we can read default value from config file. Moreover for worker pool we will use polling approach. So let's not add zmq url in docker-compose file IMO.

Ram-srini: We are not disabled sync mode yet. IMO we shoudn't hard code container service name in config file, same thing is passed as argument in the command. It will helps in QA or others to not modify the config file otherwise we have to maintain multiple config files.

listener/avalon_listener/tcs_listener.py
```diff
@@ -150,6 +156,26 @@ def parse_command_line(config, args):
...
+    # Check if listener is running in sync work order
+    if options.sync_mode:
+        is_sync = True
+    else:
+        is_sync = config["WorkloadExecution"]["sync_workload_execution"]
+
+    if options.zmq_url:
+        if not is_sync:
+            logger.warn("Option zmq_url has no effect!"
+                        "It is be supported "
```
manojgop, line 168: is -> will

listener/avalon_listener/tcs_listener.py
```diff
@@ -150,6 +156,26 @@ def parse_command_line(config, args):
...
+                        "in work order sync mode ON")
```
manojgop, line 169: in -> with

PR #532, opened 2020-06-15
Fix server error issue when unencrypted input data is sent to worker …

manju956:
…pool

- Skip encryption of inData or outData encryption keys during preprocess
  when "-" is passed as data encryption key.
- Skip decryption of inData or outData encryption keys at WPE during work order processing.
- This fixes https://jira.devtools.intel.com/browse/TCFH-980

### DIFF
--- a/common/sgx_workload/workload/ext_work_order_info_kme.cpp
+++ b/common/sgx_workload/workload/ext_work_order_info_kme.cpp
@@ -281,6 +281,16 @@ int ExtWorkOrderInfoKME::CreateWorkOrderKeyInfo(
         int count = this->in_work_order_keys.size();
         wo_key_info.in_data_keys.resize(count);
         for (int i=0; i<count; i++) {
+            std::string in_data_enc_key = \
+                ByteArrayToStr(wo_key_info.in_data_keys[i].decrypted_data);
+            // skip encrypting input data encryption key if
+            // its value is equivalent to empty or null or "-"
+            // these checks are performed again while decrypting data
+            // during work order processing at WPE
+            if (in_data_enc_key.empty() || "null" == in_data_enc_key ||
+                "-" == in_data_enc_key ) {
+                continue;
+            }
             wo_key_info.in_data_keys[i].decrypted_data = \
                 tcf::crypto::skenc::EncryptMessage(wo_key_info.sym_key,
                     wo_key_info.in_data_keys[i].decrypted_data);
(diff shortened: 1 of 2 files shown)


common/sgx_workload/workload/ext_work_order_info_kme.cpp
```diff
@@ -281,6 +281,14 @@ int ExtWorkOrderInfoKME::CreateWorkOrderKeyInfo(
         int count = this->in_work_order_keys.size();
         wo_key_info.in_data_keys.resize(count);
         for (int i=0; i<count; i++) {
+            std::string in_data_enc_key = \
+                ByteArrayToStr(wo_key_info.in_data_keys[i].decrypted_data);
+            // skip encrypting input data encryption key if
+            // its value is equivalent to empty or null or "-"
+            if (in_data_enc_key.empty() || "null" == in_data_enc_key ||
+                "-" == in_data_enc_key ) {
+                break;
```
manojgop, line 290: Please use `continue` instead of `break`.  Need to check each key in the list.

tc/sgx/trusted_worker_manager/enclave/work_order_preprocessed_keys_wpe.cpp
```diff
@@ -37,6 +37,14 @@ void WorkOrderPreProcessedKeys::Unpack(const JSON_Object* keys_object,
 
     data_keys.index = GetJsonNumber(keys_object, "index");
     encrypted_data_key = GetJsonStr(keys_object, "encrypted-data-key");
+
+    // skip decrypting data encryption key if its value is
+    // equivalent to empty or null or "-"
+    if (encrypted_data_key.empty() ||  "null" == encrypted_data_key ||
+        "-" == encrypted_data_key) {
+        return;
```
manojgop, line 45: Does `data_keys.decrypted_data` has some valid data at this point. Or should it be explicitly set it as  `data_keys.decrypted_data = Base64EncodedStringToByteArray(encrypted_data_key)`

manju956: At this point, data_keys.decrypted_key will have valid data in ByteArray form. No Base64 conversion is needed/

common/sgx_workload/workload/ext_work_order_info_kme.cpp
```diff
@@ -281,6 +281,14 @@ int ExtWorkOrderInfoKME::CreateWorkOrderKeyInfo(
         int count = this->in_work_order_keys.size();
         wo_key_info.in_data_keys.resize(count);
         for (int i=0; i<count; i++) {
+            std::string in_data_enc_key = \
+                ByteArrayToStr(wo_key_info.in_data_keys[i].decrypted_data);
+            // skip encrypting input data encryption key if
+            // its value is equivalent to empty or null or "-"
+            if (in_data_enc_key.empty() || "null" == in_data_enc_key ||
```
Ram-srini, line 288: According to spec
If this key is not provided or set to "null" or to "", the data is encrypted using encryptedSessionKey from the Work Order request

If the key value is set to "-", the data item is not encrypted, a.k.a. sent as clear text

https://entethalliance.github.io/trusted-computing/spec.html#sec-work-order-input-and-output-data-format

manju956: That's correct from Singleton perspective. At KME, while preprocessing, if dataEncryptionKey is empty or null or "-", i have to skip encrypting the data encryption key by using one time symmetric key that gets generated during preprocess. My comment is to explaining this.

Ram-srini: From KME side all("null", "", "-") are treated as same? Who is handling "null" and "" to de-crypt data using encryptedSessionKey? Is WPE doing that?

manju956: Yes, they all treated the same way. all these checks are performed again at WPE while doing actual work order processing..

Ram-srini: Please add that comment here because it slightly different from spec of singleton worker.

manju956: Done

PR #534, opened 2020-06-15
Handle negative cases during KME preprocessing

manju956:
- Handle error response during preprocessing at KME and skip actual
  work order processing at WPE

### DIFF
--- a/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
+++ b/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
@@ -28,6 +28,7 @@ from avalon_enclave_manager.worker_kv_delegate import WorkerKVDelegate
 from avalon_enclave_manager.work_order_kv_delegate import WorkOrderKVDelegate
 from listener.base_jrpc_listener import parse_bind_url
 from avalon_enclave_manager.kme.kme_listener import KMEListener
+from jsonrpc.exceptions import JSONRPCDispatchException
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -192,7 +193,12 @@ def PreProcessWorkOrder(self, **params):
             return wo_response_json["result"]
         else:
             logger.error("Could not preprocess work order at KME")
-            return wo_response_json
+            # For all negative cases, response should have an error field.
+            # Hence constructing JSONRPC error response with
+            # error code and message mapped to KME enclave response
+            err_code = wo_response_json["error"]["code"]
```
Ram-srini, line 199: What about returning wo_response_json itself? Whoever constructing validate error json then return wo_response_json. If it is not valid error jprc response then it needs to fixed at the place where it is consructed.

manju956: If we return wo_response_json as it is, the JSONRPCDispatcher will internally append "result", "id" and "json-rpc" (duplicate) parameters. This would result in increased complexity at WPE side to check whether the preprocess returned response with result (success case) or error (negative case).

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -192,7 +193,12 @@ def PreProcessWorkOrder(self, **params):
...
+            err_msg = wo_response_json["error"]["message"]
+            raise JSONRPCDispatchException(err_code, err_msg)
```
manojgop, line 201: Can we add a data field in `JSONRPCDispatchException` as "work order id"
https://github.com/hyperledger/avalon/blob/master/listener/avalon_listener/tcs_work_order_handler.py#L126. That way WPE will know for which work order ID KME returned error

manju956: Done

PR #537, opened 2020-06-16
Updated Makefiles as per modified directory structure

karthikamurthy:
Bux Fix: 

1) Build gets stuck at message “Building Avalon lmdb” and
    never comes out of the message On switching between Fabric build
    to Standalone build and then back to any docker builds on same repo

2) Standalone build is failing in the latest master with no module named 'zmq'

### DIFF
--- a/common/crypto_utils/Makefile
+++ b/common/crypto_utils/Makefile
@@ -40,8 +40,11 @@ install:
 clean:
 	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
 	rm -rf build deps dist *.egg-info
-	rm -f crypto/crypto.py crypto/crypto_wrap.cpp
-	rm -f verify_report/verify_report.py verify_report/verify_report_wrap.cpp
+	rm -f avalon_crypto_utils/crypto/crypto.py
+	rm -f avalon_crypto_utils/crypto/crypto_wrap.cpp
+	rm -f avalon_crypto_utils/verify_report/verify_report.py
+	rm -f avalon_crypto_utils/verify_report/verify_report_wrap.cpp
+
 	find . -iname '*.pyc' -delete
 	find . -iname '__pycache__' -delete
 
--- a/enclave_manager/Dockerfile
+++ b/enclave_manager/Dockerfile
@@ -215,6 +215,8 @@ RUN ldconfig \
  && echo "SGX_SSL=/opt/intel/sgxssl" >> /etc/environment
 
 COPY . /project/avalon
+# Docker build won't progress if host _dev folder not removed
+RUN rm -rf /project/avalon/tools/build/_dev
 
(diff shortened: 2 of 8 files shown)


enclave_manager/Makefile
```diff
@@ -38,10 +38,11 @@ install:
 	@echo INSTALLING WHEEL FILE =================
 	pip3 install $(WHEEL_FILE)
 
-clean:
-	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
-	rm -f avalon_enclave_manager/avalon_enclave.py avalon_enclave_manager/avalon_enclave_wrap.cpp
+clean:
```
danintel, line 41: Remove the tab character you added here

common/crypto_utils/Makefile
```diff
@@ -38,10 +38,13 @@ install:
...
 clean:
-	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
 	rm -rf build deps dist *.egg-info
-	rm -f crypto/crypto.py crypto/crypto_wrap.cpp
-	rm -f verify_report/verify_report.py verify_report/verify_report_wrap.cpp
+	rm -f avalon_crypto_utils/crypto/crypto.py
+	rm -f avalon_crypto_utils/crypto/crypto_wrap.cpp
+	rm -f avalon_crypto_utils/verify_report/verify_report.py
+	rm -f avalon_crypto_utils/verify_report/verify_report_wrap.cpp
+	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
```
manju956, line 46: Generally wheel file will be copied to deps folder in the directory which has Makefile. We are already deleting build folder before this. Then why do we need this if check?

karthikamurthy: Ya Manju , all wheel file is getting copied to dist folder not deps. So will remove the step of exclusively deleting the wheel file.

enclave_manager/Makefile
karthikamurthy: Done

common/crypto_utils/Makefile
karthikamurthy: done

enclave_manager/Dockerfile
```diff
@@ -215,6 +215,7 @@ RUN ldconfig \
  && echo "SGX_SSL=/opt/intel/sgxssl" >> /etc/environment
 
 COPY . /project/avalon
+RUN rm -rf /project/avalon/tools/build/_dev
```
Ram-srini, line 218: Add comment here why we are removing it. Mention it docker build won't progress if we already has it host.

karthikamurthy: Done

PR #540, opened 2020-06-17
Remove worker_id calculation worker details

Ram-srini:
Worker id is calculated from enclave manager

### DIFF
--- a/tests/Demo.py
+++ b/tests/Demo.py
@@ -70,7 +70,7 @@ def local_main(config):
                 input_json_obj = json.loads(input_json_str1)
                 wo_id = hex(random.randint(1, 2**64 - 1))
                 input_json_obj["params"]["workOrderId"] = wo_id
-                input_json_obj["params"]["workerId"] = worker_obj.worker_id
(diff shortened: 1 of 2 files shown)


tests/Demo.py
```diff
@@ -70,7 +70,7 @@ def local_main(config):
                 input_json_obj = json.loads(input_json_str1)
                 wo_id = hex(random.randint(1, 2**64 - 1))
                 input_json_obj["params"]["workOrderId"] = wo_id
-                input_json_obj["params"]["workerId"] = worker_obj.worker_id
+                input_json_obj["params"]["workerId"] = worker_id
```
manju956, line 73: How is global variable "worker_id" getting initialized?

Ram-srini: It is initialized in workerRetrieve() API.

manju956: Can we add a comment to specify this?

Ram-srini: Temporary fix. Manoj is removing this Demo.py file.

manju956: Ok

PR #543, opened 2020-06-17
Removed Worker API which is not enabled for Clients 

karthikamurthy:
1) Worker API removed for Client : Worker Register, Worker Update, Worker Set Status
2) Added details to Heart disease example
3) Updated the lmdb file location in clean.sh

Bug Fix:
1) Worker register API not handling negative scenarios like passing unknown or invalid parameter
2) Worker Update API not handling negative scenarios like passing unknown or invalid parameter
3) Heart disease workload computes and gives some results even when random unrelated data is passed during work order submit


examples/apps/heart_disease_eval/README.md
```diff
@@ -21,6 +21,8 @@ The command line client, `Demo.py`, runs a set of pre-built JSON requests.
 To run, follow the instructions in the "Testing" section of the
 [build document](../../../BUILD.md#testing).
 You may run the CLI demo in standalone mode or in a Docker container.
+Heart disease workload is designed to compute average heart disease risk, 
+even when invalid data is passed at in_data parametere via command prompt.
```
danintel, line 25: Change 
`passed at in_data parametere via` 
to
`passed in the in_data parameter via the`

karthikamurthy: done

PR #558, opened 2020-06-21
Handle cleanup of work order at startup

rranjan3:
- Replace wo-processing wo-worker-processing to hold list
   of work orders being actively processed
- Update wo-worker-pending to wo-worker-scheduled
- Rename wo-processed to wo-worker-processed
- Add table worker-pool to hold map of wokers in a pool
- Add calls to add entries to worker-pool mapping
- Add method to handle boot time cleanup for KME/Singleton
- Add LMDB API for search and delete

### DIFF
--- a/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
+++ b/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
@@ -24,8 +24,6 @@ import sys
 import avalon_enclave_manager.sgx_work_order_request as work_order_request
 import avalon_enclave_manager.kme.kme_enclave_info as enclave_info
 from avalon_enclave_manager.base_enclave_manager import EnclaveManager
-from avalon_enclave_manager.worker_kv_delegate import WorkerKVDelegate
-from avalon_enclave_manager.work_order_kv_delegate import WorkOrderKVDelegate
 from listener.base_jrpc_listener import parse_bind_url
 from avalon_enclave_manager.kme.kme_listener import KMEListener
 from jsonrpc.exceptions import JSONRPCDispatchException
@@ -65,14 +63,19 @@ class KeyManagementEnclaveManager(EnclaveManager):
         """
         logger.info("Executing boot time procedure")
 
+        wo_to_cleanup = []
+        if "cleanup_work_orders" in self._config["KvStorage"]:
+            wo_to_cleanup = self._config["KvStorage"]["cleanup_work_orders"]
+        self._wo_kv_delegate.cleanup_work_orders(wo_to_cleanup)
+
+        # Clear all workers mapped to this worker_id (pool). WPEs will have
+        # to register afresh to get into the pool now.
+        self._worker_kv_delegate.cleanup_pool(self._worker_id)
+
         # Add a new worker
         worker_info = EnclaveManager.create_json_worker(self, self._config)
         # Hex string read from config which is 64 characters long
-        worker_id = self._worker_id
-        self._worker_kv_delegate.add_new_worker(worker_id, worker_info)
-
-        # Cleanup wo-processing" table
-        self._wo_kv_delegate.cleanup_work_orders()
+        self._worker_kv_delegate.add_new_worker(self._worker_id, worker_info)
 
 # -------------------------------------------------------------------------
     def _execute_work_order(self, input_json_str, ext_data=""):
@@ -239,8 +242,11 @@ def main(args=None):
     parser.add_argument("--config-dir", help="configuration folder", nargs="+")
     parser.add_argument("--bind", help="KME listener url for requests to KME",
                         type=str)
-    parser.add_argument(
-        "--worker_id", help="Id of worker in plain text", type=str)
+    parser.add_argument("--worker_id",
+                        help="Id of worker in plain text", type=str)
+    parser.add_argument("--cleanup_work_orders",
+                        help="Comma-separated list of work orders to recover",
+                        type=str)
     (options, remainder) = parser.parse_known_args(args)
 
     if options.config:
@@ -260,6 +266,9 @@ def main(args=None):
         config["KMEListener"]["bind"] = options.bind
     if options.worker_id:
         config["WorkerConfig"]["worker_id"] = options.worker_id
+    if options.cleanup_work_orders:
+        config["KvStorage"]["cleanup_work_orders"] = \
+            options.cleanup_work_orders
 
     plogger.setup_loggers(config.get("Logging", {}))
     sys.stdout = plogger.stream_to_logger(
--- a/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
+++ b/enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
@@ -63,12 +63,11 @@ class KeyManagementEnclaveManager(EnclaveManager):
         """
         logger.info("Executing boot time procedure")
 
-        wo_to_cleanup = []
-        if "cleanup_work_orders" in self._config["KvStorage"]:
-            wo_to_cleanup = self._config["KvStorage"]["cleanup_work_orders"]
-        self._wo_kv_delegate.cleanup_work_orders(wo_to_cleanup)
+        # Cleanup all stale work orders for this worker which
+        # used old worker keys
+        self._wo_kv_delegate.cleanup_work_orders()
 
-        # Clear all workers mapped to this worker_id (pool). WPEs will have
+        # Clear all WPEs mapped to this worker_id (pool). WPEs will have
         # to register afresh to get into the pool now.
         self._worker_kv_delegate.cleanup_pool(self._worker_id)
 
@@ -244,9 +243,7 @@ def main(args=None):
                         type=str)
     parser.add_argument("--worker_id",
                         help="Id of worker in plain text", type=str)
-    parser.add_argument("--cleanup_work_orders",
-                        help="Comma-separated list of work orders to recover",
-                        type=str)
+
     (options, remainder) = parser.parse_known_args(args)
 
     if options.config:
@@ -266,9 +263,6 @@ def main(args=None):
         config["KMEListener"]["bind"] = options.bind
     if options.worker_id:
         config["WorkerConfig"]["worker_id"] = options.worker_id
-    if options.cleanup_work_orders:
-        config["KvStorage"]["cleanup_work_orders"] = \
-            options.cleanup_work_orders
 
     plogger.setup_loggers(config.get("Logging", {}))
     sys.stdout = plogger.stream_to_logger(
--- a/enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
+++ b/enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
@@ -30,8 +30,9 @@ class WorkOrderKVDelegate:
     specific changes in the KV storage.
     """
 
-    def __init__(self, kv_helper):
+    def __init__(self, kv_helper, worker_id):
         self._kv_helper = kv_helper
+        self._worker_id = worker_id
         # Key pair for work order receipt signing
         # This is temporary approach
         self.private_key = crypto_utils.generate_signing_keys()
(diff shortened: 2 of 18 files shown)


enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -239,8 +240,11 @@ def main(args=None):
     parser.add_argument("--config-dir", help="configuration folder", nargs="+")
     parser.add_argument("--bind", help="KME listener url for requests to KME",
                         type=str)
-    parser.add_argument(
-        "--worker_id", help="Id of worker in plain text", type=str)
+    parser.add_argument("--worker_id",
+                        help="Id of worker in plain text", type=str)
+    parser.add_argument("--cleanup_work_orders",
+                        help="Comma-separated list of work orders to recover",
+                        type=str)
```
manojgop, line 247: What is the format of work order  to be passed as argument?. If it's hex string then it is not user friendly.  Also how will uses know which work orders to pass. Clean up all work orders may be a better option.

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
```diff
@@ -51,9 +52,26 @@ def cleanup_work_orders(self, wo_ids=None):
...
+            workers = workers_in_pool.split(",")
+        else:
+            logger.info("No worker found in pool. Cleanup not required")
+            return True
+        processing_list = []
+        for worker in workers:
+            wo_id_active = self._kv_helper.get("wo-worker-active", worker)
+            # Add to processing_list, workorder ids that are in active state
+            # and are either part of wo_ids or wo_ids is None (indicating that
+            # all work-orders need to be restored and not selective).
```
manojgop, line 70: What does work order "active" state mean ? Are these work orders "scheduled" for execution but processing not yet started ? We were using a "wo-scheduled" table earlier. If you mean "active" as "scheduled" then can we use same name like "wo-worker-scheduled"?

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
```diff
@@ -102,9 +120,9 @@ def cleanup_work_orders(self, wo_ids=None):
             else:
                 logger.info("No response found for the workorder %s; " +
                             "hence placing the workorder request " +
-                            "back in wo-scheduled", wo)
-                self._kv_helper.set("wo-scheduled", wo,
-                                    WorkOrderStatus.SCHEDULED.name)
+                            "back in wo-worker-pending", wo)
+                self._kv_helper.csv_prepend("wo-worker-pending",
```
manojgop, line 124: What does "pending" state of work order mean ?

enclave_manager/avalon_enclave_manager/work_order_processor_manager.py
```diff
@@ -57,9 +60,13 @@ def _process_work_order_sync(self, process_wo_id):
         wo_id = self._kv_helper.csv_match_pop(
             "wo-worker-pending", self._worker_id, process_wo_id)
         if process_wo_id == wo_id:
-            # @TODO : Need to check if the workload_id in the work order
-            # is supported by this worker. Process only if check passes.
-            return self._process_work_order_by_id(wo_id)
+            wo_process_result = self._process_work_order_by_id(wo_id)
+
+            self._kv_helper.remove("wo-processing", wo_id)
+            # Cleanup tables that hold in-progress work orders
```
manojgop, line 66: By "in-progress" do you mean work orders which are "processing" ?

rranjan3: Yes, work orders that have been picked up from scheduled/pending to be processed but haven't yet finished processing.

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
rranjan3: "pending" implies submitted but not picked up for processing yet.

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
rranjan3: "active" is like the ones that are at some stage of processing. They are no more "pending".

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
rranjan3: This is an optional argument specifically usable when a situation is hit where we need to recover only selective work-orders.

enclave_manager/avalon_enclave_manager/kme/kme_enclave_manager.py
```diff
@@ -260,6 +264,9 @@ def main(args=None):
         config["KMEListener"]["bind"] = options.bind
     if options.worker_id:
         config["WorkerConfig"]["worker_id"] = options.worker_id
+    if options.cleanup_work_orders:
```
Ram-srini, line 267: What is this cleanup work orders parameter in config file? can you please add a comment here.

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
```diff
@@ -51,9 +52,26 @@ def cleanup_work_orders(self, wo_ids=None):
...
+        # enclave_id of a WPE).
+        workers_in_pool = self._kv_helper.get("worker-pool", self._worker_id)
+        if workers_in_pool is not None:
+            workers = workers_in_pool.split(",")
+        else:
+            logger.info("No worker found in pool. Cleanup not required")
+            return True
+        processing_list = []
+        for worker in workers:
+            wo_id_active = self._kv_helper.get("wo-worker-active", worker)
```
Ram-srini, line 67: We are creating so many tables, can you add a comment in code / document of each 
 table name and what is structure(key and value format). It is difficult to follow the code otherwise.

manojgop:
As of now since we don't have added support of recreating enclave state (Singleton and KME) from sealed storage. So in the first phase, worker restart would mean we create a new signing and encryption key pair when singleton/pinned or KME is started. That would mean we also need to clean up all work orders handled by old worker instances from the DB.

Singleton
=======
- Start Singleton 1st instance - cleanup worker and corresponding work-orders tables. If clean up all argument is passed during startup cleanup all tables (all workers and all work orders).

- Start Singleton 2,3 or more instances - by default cleanup only workers corresponding to the worker id passed and the corresponding old work orders matching the worker id. Note: cleanup all should not be ideally used which starting 2,3 singleton instances.

KME
====
KME 1, 2, 3 etc instance has to cleanup-workers and associated work orders mapped to this workers. 
We have not implemented restoring KME state from sealed data.
No restart scenario for KME. Every time KME is started a new key pair for worker is created. That means existing work orders in all the tables associated with a given worker id cannot be processed.

worker-pool -> (worker_id, list of enclave_id). for singleton enclave_id is same as worker_id
As part of KME restart clean up worker-pool remove the entry (worker_id, list of enclave_id) from the "worker-pool" table.

WPE
====
WPE 1st instance if passed with cleanup-work-orders argument can clean up all work orders corresponding to the worker id.
If cleanup-work-orders argument is not passed then work orders are not cleaned.

WPE 2,3 or higher instances by default are not required to cleanup existing work orders corresponding to the worker.

WPE instance if started with restore-work-order will move all work orders from the wo-processing tables to wo-schedule table.

Please rename below tables
wo-worker-active --> `woprocessor-wo-map`
wo-worker-pending -> `wo-worker-list`

`wo-worker-list` -> key = worker id, value = list of work order ids (csv format)
- Listener will add entry to this table when it receives work order submit request from client
- KME can cleanup the entries in this table with cleanup-workers argument during start.

`woprocessor-wo-map` -> key = enclave id, value = work order id,
in case of singleton enclave id is worker id and in case of WPE this is hex of public key of WPE encryption key
- WPE will add entry to this table by moving the work order from `wo-worker-list` table.
- WPE will also update "wo-processing" to PROCESSING just before it starts processing the work 
   order.
- WPE will remove entry from "wo-worker-processing" and "wo-processing" after it finish processing 
  the work order and then it adds entry to "wo-processed" table with status as SUCCESS/FAILED.
- WPE can clean up the entries in this table with cleanup-workorder argument during start.
  KME can cleaup the entries in this table with cleanup-workers argument during start.

`worker-pool` -> (worker_id, list of enclave_id). for singleton encalve_id is same as worker_id
- WPE will add worker_id key and enclave id as value to the list after it has successfully registered 
   with KME.
- The entry  will be removed from  worker-pool only when KME is started with cleanup-workers

enclave_manager/avalon_enclave_manager/worker_kv_delegate.py
```diff
@@ -58,6 +58,17 @@ def cleanup_worker(self, worker_ids=None):
                 result &= self._kv_helper.remove("workers", worker)
         return result
 
+    def cleanup_pool(self, worker_id):
+        """
+        Remove worker pool mapping from datastore. All workers need to
+        register afresh to be a part of the pool.
```
manojgop, line 64: `cleanup_pool()` has to be called by both KME and Singleton or only KME? I see `update_worker_map()` called by both Singleton and WPE.

enclave_manager/avalon_enclave_manager/work_order_kv_delegate.py
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import json
+import time
```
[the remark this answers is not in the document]
rranjan3: cleaned

enclave_manager/avalon_enclave_manager/worker_kv_delegate.py
rranjan3: In case of singleton, removing and adding would yield the same result.

listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -234,13 +260,18 @@ def WorkOrderSubmit(self, **params):
...
+
             # if max count reached clear a processed entry
             work_orders = self.kv_helper.lookup("wo-timestamps")
             for id in work_orders:
 
                 # If work order is processed then remove from table
-                if(self.kv_helper.get("wo-processed", id) is not None):
-                    self.kv_helper.remove("wo-processed", id)
+                if id in processed_wo_ids:
+                    self.kv_helper.csv_search_delete(
```
Ram-srini, line 273: csv_search_delete() is called for every work order request and It is O(n) operation, we might come across bottleneck here because all other workers are blocked until this completed.

manojgop: We will hit this condition only when work orders in DB is more than max work order count. So this won't be called for every work order request. For clarity, it's better to move this block to a separate function. Also (different PR) we could start a different thread for cleanup of work orders instead of doing it as part of workOrderSubmit handling

rranjan3: True @Ram-srini, that it will be O(n) and delay will be directly proportional to the number of work order id that would have been processed. But it might not be called for all work order. Rather should not be. Ii would only be so when the queue is almost topping out and that in turn would happen at the maximum possible throughput that can be churned out of the system. Having said that, still the performance at that point of time should also matter. this was chosen over introducing another table to ensure operations are as atomic as possible and refrain from multiple updates to reflect same state change.

On a side note, since its an interesting topic, I will continue on to say that the average time complexity for the operation would be in fact close to O(1). This is because a search is done on a queue which already maintains the order same as the one maintained over the items being iterated upon here. So, there is a very high chance(~99%) that the first item is hit every time in the search.

listener/avalon_listener/tcs_work_order_handler_sync.py
```diff
@@ -126,12 +126,17 @@ def WorkOrderSubmit(self, **params):
...
             for id in work_orders:
                 # If work order is processed then remove from table
-                if(self.kv_helper.get("wo-processed", id) is not None):
-                    self.kv_helper.remove("wo-processed", id)
+                if id in processed_wo_ids:
+                    self.kv_helper.csv_search_delete(
+                        "wo-worker-processed", worker_id, id)
                     self.kv_helper.remove("wo-requests", id)
                     self.kv_helper.remove("wo-responses", id)
                     self.kv_helper.remove("wo-receipts", id)
```
manojgop, line 142: Lot of duplicate code in WorkOrderSubmit() function for sync and async handling. This could be cleaned up (separate PR)

PR #561, opened 2020-06-22
Update shared_kv_store and csv_append fix

rranjan3:

### DIFF
--- a/shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
+++ b/shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
@@ -82,9 +82,10 @@ class KvDBStore(KvStorage, KvCsvStorage):
         try:
             self._db_store.db_store_put(table, key, value)
             return True
-        except Exception as ex:
-            logger.error("Exception occurred in set - {}".
-                         format(ex))
+        except Exception:
+            # @TODO : Instead of suppressing exception here, pass it back
+            # and let the caller decide how to react to the exception.
+            logger.warn("Could not set key-value in database.")
             return False
 
 # ---------------------------------------------------------------------------------------------------
@@ -101,9 +102,10 @@ class KvDBStore(KvStorage, KvCsvStorage):
                 value = self._db_store.db_store_get(table, key)
             else:
                 value = None
-        except Exception as ex:
-            logger.error("Exception occurred in get - {}".
-                         format(ex))
+        except Exception:
+            # @TODO : Instead of suppressing exception here, pass it back
+            # and let the caller decide how to react to the exception.
+            logger.warn("Could not retrieve value from database.")
             value = None
 
         if not value:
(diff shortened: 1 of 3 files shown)


shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
```diff
@@ -82,9 +82,10 @@ def set(self, table, key, value):
         try:
             self._db_store.db_store_put(table, key, value)
             return True
-        except Exception as ex:
-            logger.error("Exception occurred in set - {}".
-                         format(ex))
+        except Exception:
```
Ram-srini, line 85: In what case db_store_put will throw an error? Add that in a comment as well.

shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
```diff
@@ -101,9 +102,10 @@ def get(self, table, key):
                 value = self._db_store.db_store_get(table, key)
             else:
                 value = None
-        except Exception as ex:
-            logger.error("Exception occurred in get - {}".
-                         format(ex))
+        except Exception:
+            # @TODO : Instead of suppressing exception here, pass it back
```
Ram-srini, line 106: Is db_store_get throw exception if key does not exists? If Yes then add comment.

shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
```diff
@@ -132,9 +134,10 @@ def remove(self, table, key, value=None):
             else:
                 self._db_store.db_store_del(table, key, value)
             return True
-        except Exception as ex:
-            logger.error("Exception occurred in remove - {}".
-                         format(ex))
+        except Exception:
+            # @TODO : Instead of suppressing exception here, pass it back
```
Ram-srini, line 138: same here, add a comment when this throws an exception.

shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
rranjan3: @Ram-srini, added TODO for now. Need to put up a full fledged error handling here. As of now several error situation are handled in this blanket handle. The varied errors could be - failing to open database in the environment, failing to create a transaction, failing to run the transaction or if there is a size mismatch between the size of value retrieved/to be inserted and the size calculated.
Will take up the changes in a separate PR to create new error codes and pass them back to caller for them to handle appropriately.

Ram-srini: Good add error code check in caller side/throw exception in case set fails.

shared_kv_storage/kv_storage/remote_lmdb/shared_kv_dbstore.py
```diff
@@ -82,9 +82,10 @@ def set(self, table, key, value):
         try:
             self._db_store.db_store_put(table, key, value)
             return True
-        except Exception as ex:
-            logger.error("Exception occurred in set - {}".
-                         format(ex))
+        except Exception:
+            # @TODO : Instead of suppressing exception here, pass it back
+            # and let the caller decide how to react to the exception.
+            logger.warn("Could not set key-value in database.")
```
manju956, line 88: Change log level to error in case of exception

manojgop: @manju956 This may not be an error case always. The caller of the API will have the right context to decide if this is an error case.

PR #568, opened 2020-06-24
Updating the document and code correction

karthikamurthy:
Updating the BUILD.md and PREREQUISITES.md as per new tcs_config.toml 
Code correction in base_jrpc_listener.py to send back proper response to client

### DIFF
--- a/common/python/Makefile
+++ b/common/python/Makefile
@@ -35,7 +35,7 @@ install:
 	pip3 install $(WHEEL_FILE)
 
 clean:
-	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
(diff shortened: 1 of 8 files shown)


common/python/Makefile
```diff
@@ -35,7 +35,7 @@ install:
 	pip3 install $(WHEEL_FILE)
 
 clean:
-	if [ -f $(WHEEL_FILE) ] ; then pip3 uninstall --yes $(WHEEL_FILE); fi
+	if pip3 uninstall --yes $(WHEEL_FILE); then echo UNINSTALLED WHEEL FILE ; fi
```
manju956, line 38: Can we print which WHEEL file got uninstalled?
Applicable to all such occurances.

karthikamurthy: done

PR #586, opened 2020-06-29
Cleanup enclave_type enum from request handler

manju956:
- Since enclave binaries for different workers are already in place,
  there is no need of enclave type identifier, hence deleted all
  references.
- Exception handling done in Singleton and WPE enclave managers


enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
```diff
@@ -32,15 +32,12 @@ def __init__(self, config, work_order, ext_data=""):
...
                 "avalon_enclave_manager.kme.kme_enclave")
-            self.enclave_type = self.enclave.KME_ENCLAVE
         elif enclave_type == "WPE":
             self.enclave = importlib.import_module(
                 "avalon_enclave_manager.wpe.wpe_enclave")
-            self.enclave_type = self.enclave.WPE_ENCLAVE
         else:
             self.enclave = importlib.import_module(
                 "avalon_enclave_manager.singleton.singleton_enclave")
-            self.enclave_type = self.enclave.SINGLETON_ENCLAVE
```
Ram-srini, line 43: Instead of else can you use specific name for singleton like "pinned_enclave". In toml if specify any value then it will go to singleton. Validation team has bug filed for this.

manju956: Makes sense. Will add a check for singleton enclave as well.

PR #587, opened 2020-06-30
Proper error messages added for empty WO request.

karthikamurthy:
Proper error messages added for empty requester nonce and empty worker encryption key

Bug Fix:
1)Incorrect error "Invalid Params" appear on submiting WO request with empty requester noonce/worker encryption key params

Signed-off-by: Karthika Murthy <karthika.murthy@intel.com>

### DIFF
--- a/examples/apps/generic_client/eth_generic_client.py
+++ b/examples/apps/generic_client/eth_generic_client.py
@@ -189,21 +189,27 @@ def _create_work_order_params(worker_id, workload_id, in_data,
     requester_id = secrets.token_hex(32)
     requester_nonce = secrets.token_hex(16)
     # Create work order params
-    wo_params = WorkOrderParams(
-        work_order_id, worker_id, workload_id, requester_id,
-        session_key, session_iv, requester_nonce,
-        result_uri=" ", notify_uri=" ",
-        worker_encryption_key=worker_encrypt_key,
-        data_encryption_algorithm="AES-GCM-256"
-    )
+    try:
+        wo_params = WorkOrderParams(
+            work_order_id, worker_id, workload_id, requester_id,
+            session_key, session_iv, requester_nonce,
+            result_uri=" ", notify_uri=" ",
+            worker_encryption_key=worker_encrypt_key,
+            data_encryption_algorithm="AES-GCM-256"
+        )
+    except Exception as err:
+        return False, err
+
     # Add worker input data
     for value in in_data:
         wo_params.add_in_data(value)
 
     # Encrypt work order request hash
-    wo_params.add_encrypted_request_hash()
+    code, out_json = wo_params.add_encrypted_request_hash()
+    if not code:
+        return code, out_json
 
-    return wo_params
+    return True, wo_params
 
 
 def _create_work_order_receipt(wo_receipt, wo_params,
@@ -489,9 +495,13 @@ def Main(args=None):
 
     # Create work order
     verification_key = worker_obj.verification_key
-    wo_params = _create_work_order_params(worker_id, workload_id,
-                                          in_data, worker_obj.encryption_key,
-                                          session_key, session_iv)
+    code, wo_params = _create_work_order_params(
+                            worker_id, workload_id,
+                            in_data, worker_obj.encryption_key,
+                            session_key, session_iv)
+    if not code:
+        logger.error("Work order submission failed")
+        exit(1)
 
     client_private_key = crypto_utility.generate_signing_keys()
     if requester_signature:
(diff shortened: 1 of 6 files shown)


examples/apps/generic_client/eth_generic_client.py
```diff
@@ -489,9 +491,14 @@ def Main(args=None):
...
-    wo_params = _create_work_order_params(worker_id, workload_id,
-                                          in_data, worker_obj.encryption_key,
-                                          session_key, session_iv)
+    code, wo_params = _create_work_order_params(
+                            worker_id, workload_id,
+                            in_data, worker_obj.encryption_key,
+                            session_key, session_iv)
+    if not code:
+        logger.error("Work order submission failed")
+        logger.error(wo_params)
```
manju956, line 500: log wo_params in the previous log only.

sdk/avalon_sdk/work_order/work_order_params.py
```diff
@@ -41,17 +42,17 @@ def __init__(
...
                 worker_encryption_key.encode("UTF-8").hex())
         if data_encryption_algorithm:
             self.set_data_encryption_algorithm(data_encryption_algorithm)
-
-        encrypted_session_key = crypto_utility.generate_encrypted_key(
-            session_key, worker_encryption_key)
-        self.set_encrypted_session_key(
-            crypto_utility.byte_array_to_hex(encrypted_session_key)
-        )
+        if worker_encryption_key and worker_encryption_key.strip() and \
```
manojgop, line 50: Why do we need `worker_encryption_key.strip()` and `session_key.strip()` ? Ideally there should not be any leading/trailing spaces in worker_encryption_key or session_key.  If it's there then it's an error case

karthikamurthy: yes, blank spaces when passed in worker_encryption_key is not considered as None and hence  the code breaks at crypto utility function. Its a negative case which needs to be filtered during parameter check itself.

examples/apps/generic_client/eth_generic_client.py
karthikamurthy: thanks Manju, Second log is not required. It will be removed.

sdk/avalon_sdk/work_order/work_order_params.py
manojgop: We should not filter it. If blank spaces are passed either check if it's a valid hex string or let the crypto fail and return crypto error.

manojgop: We should not filter it and remove spaces. If blank spaces or any invalid characters are passed are passed either check if it's a valid hex string or let the crypto fail and return crypto error.

karthikamurthy: strip function is removed

sdk/avalon_sdk/work_order/work_order_params.py
```diff
@@ -239,6 +248,8 @@ def get_out_data(self):
 
     def get_requester_nonce(self):
         """Return requesterNonce work order parameter."""
+        if not (self.params_obj["requesterNonce"]).strip():
```
Ram-srini, line 251: Good to add this check in set rather in get call.

karthikamurthy: done

sdk/avalon_sdk/work_order/work_order_params.py
```diff
@@ -42,16 +44,19 @@ def __init__(
         self.set_workload_id(workload_id)
         self.set_requester_id(requester_id)
         if worker_encryption_key:
-            self.set_worker_encryption_key(
-                worker_encryption_key.encode("UTF-8").hex())
+                self.set_worker_encryption_key(
+                    worker_encryption_key)
         if data_encryption_algorithm:
```
manju956, line 49: Can we add validation for data_encryption_algorithm? It should be a string value. 
QA team raised this concern, if they pass list of algorithms from generic client, it was still going fine.

sdk/avalon_sdk/work_order/work_order_params.py
```diff
@@ -115,37 +124,48 @@ def set_session_key_iv(self, session_iv):
...
+            self.final_hash = crypto_utility.compute_message_hash(concat_hash)
+            encrypted_request_hash = crypto_utility.encrypt_data(
+                self.final_hash, self.session_key, self.session_iv)
+            encrypted_request_hash_hex = crypto_utility.byte_array_to_hex(
+                                            encrypted_request_hash)
+            self.params_obj["encryptedRequestHash"] = \
+                encrypted_request_hash_hex
+            return True, ""
+
+        except Exception as err:
```
rranjan3, line 164: Did we really want to pass a tuple here? As the second value i.e.- the error response is always fixed. returning True/False would have sufficed.

sdk/avalon_sdk/work_order/work_order_params.py
karthikamurthy: done

PR #593, opened 2020-07-01
Change behavior of work get result for fabric

Ram-srini:
First make a call to chaincode query function to get
the work order result. If result not found in blockchain
then wait for workOrderCompleted event


sdk/avalon_sdk/connector/blockchains/fabric/fabric_work_order.py
```diff
@@ -138,6 +137,20 @@ def handle_fabric_event(event, block_num, txn_id, status):
...
+        if (self.__fabric_wrapper is not None):
+            params = []
+            params.append(work_order_id)
+            txn_status = self.__fabric_wrapper.invoke_chaincode(
+                self.CHAIN_CODE,
+                'workOrderGet',
+                params)
+            # workOrderGet returns tuple containing work order
+            # result params as defined in EEA spec 6.10.5
+            if txn_status and len(txn_status) > 0:
```
rranjan3, line 149: This should take care of invalid workorder id as well. Isn't it? Or is it already handled?

Ram-srini: added timeout for asyncio event loop. that fix invalid work order id

PR #595, opened 2020-07-02
Remove openssl dependency from crypto_utils

manju956:
- Remove openssl dependency from crypto_utils
- Create separate python package for verify_report
- Delete verify report from crypto_utils_openssl


common/crypto_utils/avalon_crypto_utils/crypto/crypto.i
```diff
@@ -18,7 +18,6 @@
 %{
```
manojgop, line 18: We don't need SWIG in avalon_crypto_utils. Only verify report need SWIG. So you can delete this file

common/crypto_utils/setup.py
```diff
@@ -58,59 +42,31 @@
 
 crypto_include_dirs = [
     os.path.join(tcf_root_dir, 'common/cpp'),
```
manojgop, line 44: avalon_crypto_utils uses Pycrytodomex and is not dependent on cpp.

common/crypto_utils/avalon_crypto_utils/crypto/crypto.i
manju956: Ok, then crypto folder itself can be removed.

manju956: Removing swig file needs lot of changes in listener and other files. We have to provide python alternatives for byte_array_to_str, random_bit_string etc.. To keep this PR simple, i will take this change in a separate pr.

manojgop: Let's add the required functions first (in a new PR). We can put this PR on hold till then. I don't see a point why we need to merge this PR till we avoid openSSL dependency completely in crypto utils.

manju956: I created a new commit in the same PR to provide python definitions for common/cpp utils as alternate.

common/crypto_utils/setup.py
manju956: Done

common/crypto_utils/avalon_crypto_utils/keys.py
```diff
@@ -35,13 +36,13 @@ def __init__(self, verifying_key, encryption_key):
...
-        self._verifying_key = crypto.SIG_PublicKey(verifying_key)
-        self._encryption_key = crypto.PKENC_PublicKey(encryption_key)
+        self._verifying_key = VerifyingKey.from_pem(verifying_key)
+        self._encryption_key = RSA.importKey(encryption_key)
 
     # -------------------------------------------------------
     @property
     def identity(self):
-        return self._verifying_key.Serialize()
+        return self._verifying_key.to_string()
```
manojgop, line 45: Does this give PEM formatted string. Else you can use `self._verifying_key.to_pem()`

common/crypto_utils/Makefile
```diff
@@ -12,11 +12,13 @@
...
 # limitations under the License.
 
-PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
+#PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | cut -b 1}
 MOD_VERSION=${shell ../../bin/get_version}
 
 # Format of wheel package is <pkg_name>-{python-tag}-{abi-tag}-{platform}
-WHEEL_FILE=dist/avalon_crypto_utils-${MOD_VERSION}-cp${PY_VERSION}-cp${PY_VERSION}m-linux_x86_64.whl
+WHEEL_FILE=dist/avalon_crypto_utils-${MOD_VERSION}-py${PY_VERSION}-none-any.whl
```
manojgop, line 20: Please remove below lines in `.clean` section
```
rm -f avalon_crypto_utils/crypto/crypto.py
rm -f avalon_crypto_utils/crypto/crypto_wrap.cpp
rm -f avalon_crypto_utils/verify_report/verify_report.py
rm -f avalon_crypto_utils/verify_report/verify_report_wrap.cpp
```

common/crypto_utils/Makefile
```diff
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
+#PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | tr -d .}
+PY_VERSION=${shell python3 --version | sed 's/Python \(3\.[0-9]\).*/\1/' | cut -b 1}
```
manojgop, line 16: Were you facing any issues with `tr -d `

common/crypto_utils/setup.py
```diff
@@ -32,85 +32,13 @@
...
-    'uavalon-common',
-    'uavalon-base64',
-    'uavalon-parson',
-] + openssl_libs
-
-crypto_modulefiles = [
-    "avalon_crypto_utils/crypto/crypto.i",
-]
-
-crypto_module = Extension(
```
manojgop, line 91: import of `Extension` at top can also be removed

enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, enclave_type, work_order, ext_data=""):
 
     # Execute work order in Intel SGX worker enclave
     def execute(self):
-        serialized_byte_array = crypto.string_to_byte_array(self.work_order)
+        serialized_byte_array = crypto.str_to_byte_array(self.work_order)
```
manojgop, line 46: you could have kept the function name exactly the same ? Is this the only occurrence of `string_to_byte_array`

listener/avalon_listener/tcs_workorder_receipt_handler.py
```diff
@@ -17,7 +17,6 @@
 import base64
 from utility.hex_utils import is_valid_hex_str
 import avalon_crypto_utils.crypto_utility as crypto_utility
-import avalon_crypto_utils.crypto.crypto as crypto
 from error_code.error_status import ReceiptCreateStatus, SignatureStatus,\
```
manojgop, line 20: Need to remove openSSL and common cpp build from listener Dockerfile -> https://github.com/hyperledger/avalon/blob/master/listener/Dockerfile#L99

common/crypto_utils/avalon_crypto_utils/keys.py
```diff
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 import hashlib
-import avalon_crypto_utils.crypto.crypto as crypto
+from ecdsa import VerifyingKey
+from Cryptodome.PublicKey import RSA
```
manojgop, line 17: Enclave Manager this not using the EnclaveKeys. Object  is getting initialzied but seems it's not used. For example 
https://github.com/hyperledger/avalon/blob/master/enclave_manager/avalon_enclave_manager/singleton/singleton_enclave_info.py#L61

You could remove the occurrence of `EnclaveKeys` and also remove this file `keys.py`

common/crypto_utils/Makefile
manju956: Now since the crypto_utils package is independent of swig, the python tag in package name will only have "py3" as opposed to "py36" with swig. Cut is used to remove everything after 3, in the python_version.
However earlier trim was used to remove the dot between python version 3.6.

Same pattern is used to determine PY_VERSION here as well - https://github.com/hyperledger/avalon/blob/master/common/python/Makefile#L16

enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
manju956: Ok, will rename to `string_to_byte_array`.This is the only place which is used.

common/crypto_utils/avalon_crypto_utils/keys.py
manju956: Done

common/crypto_utils/Makefile
manju956: Done

common/crypto_utils/setup.py
manju956: Removed

enclave_manager/avalon_enclave_manager/sgx_work_order_request.py
manju956: Done

listener/avalon_listener/tcs_workorder_receipt_handler.py
manju956: Done

common/crypto_utils/avalon_crypto_utils/keys.py
manju956: Done

common/crypto_utils/Makefile
manojgop: Please remove commented line

manju956: Done

PR #597, opened 2020-07-02
Compute work order request and response hash as per Trusted compute s…

manojgop:
…pec.

Current implementation of Request and response hash computation has
base64 conversions which is not as per TC spec. This commit fixes this issue.


common/crypto_utils/avalon_crypto_utils/crypto_utility.py
```diff
@@ -252,10 +252,10 @@ def byte_array_to_hex(byte_array):
 
 
 # -----------------------------------------------------------------------------
-def compute_message_hash(message):
+def compute_message_hash(message_bytes):
```
danintel, line 255: As long as you are changing this, add a docstring. Something like this (modified from crypto_utils.cpp):

```
"""
Compute SHA256 hash of message_bytes.

Parameters:
    message_bytes Data to hash
Returns:
    binary hash of data
"""
```

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
```diff
@@ -308,37 +308,56 @@ namespace tcf {
...
-        ByteArray hash_data;
-        std::string hash_2;
+        ByteArray hash_1 =  tcf::crypto::ComputeMessageHash(
+                                    StrToByteArray(concat_string));
+        // Calculate inData hash
+        ByteArray hash_in_data;
+        std::string hash_in_data_str;
         size_t i;
+        // First sort the inData elements based on index
+        // Sorting is requried to calculate hash deterministically
```
danintel, line 318: s/requried/required/

common/crypto_utils/avalon_crypto_utils/crypto_utility.py
manojgop: Done

tc/sgx/trusted_worker_manager/enclave/work_order_processor.cpp
manojgop: Done

PR #599, opened 2020-07-02
Remove command line workload configuration

manju956:
- Since build time configuration is added to configure workloads,
  command line opion is no longer needed.

### DIFF
--- a/config/singleton_enclave_config.toml
+++ b/config/singleton_enclave_config.toml
@@ -76,11 +76,7 @@ enclave_library_path = "tc/sgx/trusted_worker_manager/enclave/build/lib/"
 # Id of worker in plain text. This can be overridden by passing values in command line for
 # enclave manager using --worker_id option.
 worker_id = "singleton-worker-1"
-# Supported workloads for this worker. The work orders matching these workload ids
-# will be processed by the enclave manager. The enclaves should be built with matching
-# workloads. This can be overridden by passing values in command line for enclave manager
-# using --workloads option.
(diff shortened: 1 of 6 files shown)


config/singleton_enclave_config.toml
```diff
@@ -76,11 +76,7 @@ enclave_library_path = "tc/sgx/trusted_worker_manager/enclave/build/lib/"
 # Id of worker in plain text. This can be overridden by passing values in command line for
 # enclave manager using --worker_id option.
 worker_id = "singleton-worker-1"
-# Supported workloads for this worker. The work orders matching these workload ids
-# will be processed by the enclave manager. The enclaves should be built with matching
-# workloads. This can be overridden by passing values in command line for enclave manager
-# using --workloads option.
-workloads = "echo-result,heart-disease-eval,inside-out-eval,simple-wallet"
```
manojgop, line 83: My understanding is this is still needed. Using this configuration enclave managers knows which workloads it needs to read from LMDB to process

rranjan3: This is not the case @manojgop. The work orders are picked up from dbstore solely on the basis of worker-id as of now.
If you imply for future use, it can be pondered upon. Going ahead when the dbstore has more context of a workload and workers(implying WPEs here) themselves, then it would make sense. As it stands today, we need to pry open each work order to get to know what workload is in there.

manojgop: It will be useful for WPE (enclave) to provide an interface which WPE Manager can use to fetch the list of supported workloads. This can be done in a separate PR

PR #603, opened 2020-07-04
Refactor transaction build/execute in Ethereum

rranjan3:
- Create a common method to build and execute transaction
 - Retry if an error is encountered in trnsaction commit as
   these error are many a times indiative of a transaction
   not being mined within the timeout period. Then one of
   the most recurrent reason for not being mined is a clash
   in the nonce for transactions. This clash is being taken
   care now by introducing a random sleep.
 - Transaction timeout decreased from default 120s to 10s.
   This being a private network with minimum gas required
   being 0, there should not be such huge delay in mining.
 - Poll latency for ethereum transactions updated to 0.2s
   to cut down excessive traffic when load is higher.
 - Decreasing blockperiod for default Besu network

### DIFF
--- a/sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
+++ b/sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
@@ -20,6 +20,8 @@ from solcx import compile_source, set_solc_version, get_solc_version
 from urllib.parse import urlparse
 import web3
 import json
+import random
+import time
 
 logging.basicConfig(
     format='%(asctime)s - %(levelname)s - %(message)s', level=logging.INFO)
@@ -189,8 +191,12 @@ class EthereumWrapper():
         signed_tx = self.__w3.eth.account.signTransaction(
             tx_dict, private_key=self.__eth_private_key)
         tx_hash = self.__w3.eth.sendRawTransaction(signed_tx.rawTransaction)
+        # This being a private network with minimum gas required being 0,
+        # there should not be a huge delay in mining. Hence timeout is 10s.
+        # Poll latency is kept 0.2 (not default 0.1) to cut down excessive
+        # traffic when load is higher.
         tx_receipt = self.__w3.eth.waitForTransactionReceipt(
-            tx_hash.hex(), 120)
+            tx_hash.hex(), timeout=10, poll_latency=0.2)
         logging.debug("Executed transaction hash: %s, receipt: %s",
                       format(tx_hash.hex()), format(tx_receipt))
         return tx_receipt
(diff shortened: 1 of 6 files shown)


sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
```diff
@@ -207,11 +209,49 @@ def execute_unsigned_transaction(self, tx_dict):
...
+                logging.warn("Exception in writing transaction to"
+                             " blockchain: {}".format(ex))
+                # Retry 3 times only
+                if retry_count == 3:
+                    if isinstance(ex, web3.exceptions.TimeExhausted):
+                        raise TimeoutError(
+                            "Transactions are being submitted at a faster "
+                            "rate than they can be processed.")
+                    else:
+                        raise
```
Ram-srini, line 253: Is raise Exception type missing..

sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
```diff
@@ -190,7 +192,7 @@ def sign_execute_raw_transaction(self, tx_dict):
             tx_dict, private_key=self.__eth_private_key)
         tx_hash = self.__w3.eth.sendRawTransaction(signed_tx.rawTransaction)
         tx_receipt = self.__w3.eth.waitForTransactionReceipt(
-            tx_hash.hex(), 120)
+            tx_hash.hex(), timeout=10, poll_latency=0.2)
```
Ram-srini, line 195: Add comment about parameter poll_latency.

sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
rranjan3: No `raise` is like a pass through when used alone. It will forward the exception received.
Just for understanding for all. :)
```
>>> try:
...     raise TimeoutError("Timeout")
... except Exception as ex:
...     raise
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
TimeoutError: Timeout
>>> try:
...     raise TimeoutError("Timeout")
... except Exception as ex:
...     raise Exception("Some exception occurred")
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
**TimeoutError: Timeout**

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
**Exception: Some exception occurred**
>>>
```

sdk/avalon_sdk/connector/blockchains/ethereum/ethereum_wrapper.py
rranjan3: Yes will add. Added in the commit log. Forgot here.

PR #608, opened 2020-07-05
Cleanup from all workers if processing limit hit

rranjan3:
- Cleanup work orders from all workers in the system if
   the max_work_order limit is hit
 - Boost max_work_order limit to 1000 in listener config.
   10 is way too low.

### DIFF
--- a/listener/avalon_listener/tcs_work_order_handler.py
+++ b/listener/avalon_listener/tcs_work_order_handler.py
@@ -259,26 +259,37 @@ class TCSWorkOrderHandler:
             )
 
         if((self.workorder_count + 1) > self.max_workorder_count):
-
-            # wo_ids is a csv of work order ids retrieved from database
-            wo_ids = self.kv_helper.get("wo-worker-processed", worker_id)
-            processed_wo_ids = [] if wo_ids is None else wo_ids.split(",")
+            # Lookup all workers.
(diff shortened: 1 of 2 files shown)


listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -259,26 +259,37 @@ def WorkOrderSubmit(self, **params):
...
 
             # if max count reached clear a processed entry
             work_orders = self.kv_helper.lookup("wo-timestamps")
             for id in work_orders:
-
                 # If work order is processed then remove from table
                 if id in processed_wo_ids:
                     self.kv_helper.csv_search_delete(
-                        "wo-worker-processed", worker_id, id)
+                        "wo-worker-processed", wo_worker_map[id], id)
```
Ram-srini, line 281: Remove it from map wo_worker_map as well, otherwise this map will grow very big.

listener/listener_config.toml
```diff
@@ -25,7 +25,7 @@ LogFile  = "__screen__"
 
 [Listener]
 bind = "http://localhost:1947"
-max_work_order_count = 10
+max_work_order_count = 1000
```
Ram-srini, line 28: Add description of this parameter in the config.

rranjan3: Done.

PR #616, opened 2020-07-09
Add/Enhance function comments

manju956:


tc/sgx/trusted_worker_manager/enclave/kme/ext_work_order_info_kme.h
```diff
@@ -180,9 +184,21 @@ class ExtWorkOrderInfoKME : public ExtWorkOrderInfoImpl {
...
+    /*
+     * Calculates hash on the encrypted keys, input-data-keys and output-data-keys
+     * @param wo_key_info - Instance of WorkOrderKeyInfo
+     * @param wo_key_info_hash [OUT] - Computed hash value
+    */
     void CalculateWorkOrderKeyInfoHash(WorkOrderKeyInfo wo_key_info,
         ByteArray& wo_key_info_hash);
 
+    /*
+     * Creates JSON document of encryption keys, signing keys
```
rranjan3, line 197: Did you intend to say "serialized JSON object" by "JSON document"?

tc/sgx/trusted_worker_manager/enclave/kme/ext_work_order_info_kme.h
```diff
@@ -180,9 +184,21 @@ class ExtWorkOrderInfoKME : public ExtWorkOrderInfoImpl {
...
+     * and data encryption keys
```
rranjan3, line 198: Also, why do we put the here and not where the function is defined or maybe declared (in the header)?

tc/sgx/trusted_worker_manager/enclave/kme/ext_work_order_info_kme.h
manju956: Yes, will rephrase accordingly

tc/sgx/trusted_worker_manager/enclave/kme/ext_work_order_info_kme.h
manju956: The reason to put the function comments in header file is to have function comments consolidated in one place. Now, if all the function comments in this header are moved to cpp file, the size of the source file would grow very large. Hence keeping all the comments in the header itself

danintel: One can argue for having comments in the header file or source file.

But reducing the . cpp source file size is not a consideration. We have gigabytes of storage and compilers are very fast processing source.  Also the comments processed several more times in header files than .cpp source files.

manojgop: We can keep  function comments at single place either header file or .cpp file. In this PR my understanding is the comments were move from .cpp file header file

manju956: > One can argue for having comments in the header file or source file.
> 
> But reducing the . cpp source file size is not a consideration. We have gigabytes of storage and compilers are very fast processing source. Also the comments processed several more times in header files than .cpp source files.

Agree @danintel - The point mentioned regarding size of source file is with respect to code readability.
If all comments are put inside cpp file(specifically in this file function comments are more descriptive), the readability of the file becomes bit difficult, hence moved to header file.

PR #600, opened 2020-07-02
Jsonschema library added

karthikamurthy:
Add jsonschema validation for all worker and work order request.

Signed-off-by: Karthika Murthy <karthika.murthy@intel.com>

PR #619, opened 2020-07-09
Cleanup in ethereum connector service

Ram-srini:

### DIFF
--- a/listener/avalon_listener/tcs_work_order_handler.py
+++ b/listener/avalon_listener/tcs_work_order_handler.py
@@ -206,7 +206,6 @@ class TCSWorkOrderHandler:
               as defined in EEA spec 6.1.1
         Returns jrpc response as defined in EEA spec 6.1.3
         """
-
         wo_id = params["workOrderId"]
         input_json_str = params["raw"]
         input_value_json = json.loads(input_json_str)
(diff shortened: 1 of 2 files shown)


listener/avalon_listener/tcs_work_order_handler.py
```diff
@@ -206,6 +206,11 @@ def WorkOrderSubmit(self, **params):
               as defined in EEA spec 6.1.1
         Returns jrpc response as defined in EEA spec 6.1.3
         """
+        if "workOrderId" not in params:
+            raise JSONRPCDispatchException(
+                WorkOrderStatus.INVALID_PARAMETER_FORMAT_OR_VALUE,
+                "Missing work order id"
+            )
```
manojgop, line 213: Need a similar check in https://github.com/hyperledger/avalon/blob/master/listener/avalon_listener/tcs_work_order_handler_sync.py#L65. 
But rather we should have a common function for both files

Ram-srini: This could be fixed using jsonschema as well.

rranjan3:
Depends on #600

listener/avalon_listener/tcs_work_order_handler.py
Ram-srini: @manojgop Removed since json-schema address it.

PR #620, opened 2020-07-09
Make Worker id config mandatory field

manju956:

### DIFF
--- a/docker/k8s/fabric/README.md
+++ b/docker/k8s/fabric/README.md
@@ -61,8 +61,14 @@ This serves as a playground to setup k8s cluster for Avalon Fabric proxy model.
 3. To execute a transaction from shell
    ```bash
    cd examples/apps/generic_client
-   ./fabric_generic_client.py -b fabric --workload_id "echo-result" --in_data "Hello"
+   ./fabric_generic_client.py -b fabric --workload_id "echo-result" --in_data "Hello" \
+       --worker_id "singleton-worker-1"
    ```
+
+   NOTE: `worker_id` should match with worker id of singleton enclave manager.  
+   This worker_id can either be command line argument passed to enclave manager or   
+   in the absence of command line argument, worker_id in
+   `$TCF_HOME/config/singleton_enclave_config.toml` should be used.
 
 # Heads up
 - All the resources specified in yaml are using the images built locally.
--- a/examples/apps/echo/client/echo_client.py
+++ b/examples/apps/echo/client/echo_client.py
@@ -74,10 +74,17 @@ def ParseCommandLine(args):
         help="skip URI lookup and use the registry in the config file",
         action="store_true")
 
-    parser.add_argument(
-        "-w", "--worker-id",
-        help="skip worker lookup and retrieve specified worker",
+    mutually_excl_group_worker = parser.add_mutually_exclusive_group(
+        required=True)
+    mutually_excl_group_worker.add_argument(
+        "-w", "--worker_id",
+        help="worker id in plain text to use to submit a work order",
         type=str)
+    mutually_excl_group_worker.add_argument(
+        "-wx", "--worker_id_hex",
+        help="worker id as hex string to use to submit a work order",
+        type=str)
+
     parser.add_argument(
         "-m", "--message",
         help='text message to be included in the JSON request payload',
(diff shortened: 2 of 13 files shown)


examples/apps/generic_client/generic_client.py
```diff
@@ -74,7 +74,7 @@ def _parse_command_line(args):
         default="registry",
         choices={"registry", "listing"},
         type=str)
-    mutually_excl_group_worker = parser.add_mutually_exclusive_group()
+    mutually_excl_group_worker = parser.add_mutually_exclusive_group(required=True)
```
rranjan3, line 77: Should be required in other example clients as well.

docker-compose.yaml
```diff
@@ -63,7 +63,7 @@ services:
       - 5555
     command: |
       bash -c "
-        enclave_manager --lmdb_url http://avalon-lmdb:9090
+        enclave_manager --lmdb_url http://avalon-lmdb:9090 --worker_id singleton-worker-1
```
rranjan3, line 66: Need not mention explicitly. Neither in any of the other docker-compose files. They will pick up the default. Need to pass here as well when intention is to override the one in the enclave manager config file. Rather update in the single manager config toml. That would suffice.

rranjan3:
This would also require all the documentation files to be updated -
 - All clients
 - Ethereum Besu/Ganche
 - Fabric

docker-compose.yaml
manju956: Makes sense to put it in config file as it would avoid changes at many places. Will do.

docker/k8s/fabric/README.md
```diff
@@ -61,7 +61,8 @@ This serves as a playground to setup k8s cluster for Avalon Fabric proxy model.
 3. To execute a transaction from shell
    ```bash
    cd examples/apps/generic_client
-   ./fabric_generic_client.py -b fabric --workload_id "echo-result" --in_data "Hello"
+   ./fabric_generic_client.py -b fabric --workload_id "echo-result" --in_data "Hello" \
+       --worker_id "singleton-worker-1"
```
Ram-srini, line 65: Update document that how are we getting this worker id as "singleton-worker-1". Otherwise it appears to be some random string. Update in all the places.

examples/apps/echo/client/echo_client.py
```diff
@@ -124,7 +131,13 @@ def ParseCommandLine(args):
 
     requester_signature = options.requester_signature
     input_data_hash = options.data_hash
+    # worker id
     worker_id = options.worker_id
+    worker_id_hex = options.worker_id_hex
+
+    worker_id = worker_id_hex if not worker_id \
+        else hex_utils.get_worker_id_from_name(worker_id)
+
```
Ram-srini, line 140: It would be good if we validate the worker id passed in the command line is correct or not?

manju956: Worker_id can be any name. We don't have anything to compare against.

docker/k8s/fabric/README.md
manju956: Ok, i can mention worker_id is same as the one which is configured in config toml file.

docker/k8s/fabric/README.md
```diff
@@ -61,8 +61,10 @@ This serves as a playground to setup k8s cluster for Avalon Fabric proxy model.
...
    ```
+   NOTE: `worker_id` is the same value passed in `$TCF_HOME/config/singleton_enclave_config.toml`
```
manojgop, line 67: It's not necessary to keep worker id same as that in config file. You need to mention that worker id can either be passed as a command line parameter and if worker id is not passed in command line then worker id value in config toml file will be used. Command line parameter takes precedence over that mentioned in config file.

The generic client needs to choose a worker id which matches with worker id of singleton enclave manager worker

tools/run_tests.sh
```diff
@@ -28,6 +28,7 @@ generic_client_path="${TCF_HOME}/examples/apps/generic_client"
 # Read Listener port from config file
 listener_port=`grep listener_port ${TCF_HOME}/config/tcs_config.toml | awk {'print $3'}`
 LISTENER_URL="localhost"
+WORKER_ID="singleton-worker-1"
```
manojgop, line 31: Please add a comment that this value should match with singleton enclave manager worker id

docker/k8s/fabric/README.md
manju956: Done

tools/run_tests.sh
manju956: Done

PR #622, opened 2020-07-10
Syncup workers from kv to proxy blockchain

rranjan3:
- Poll the kv periodically (30s) and sync up all workers
   to proxy blockchain. This would ensure one way syncup
   for dynamic addition of workers.

### DIFF
--- a/blockchain_connector/common/connector_common/base_connector.py
+++ b/blockchain_connector/common/connector_common/base_connector.py
@@ -31,6 +31,9 @@ from connector_common.work_order_delegate import \
 logging.basicConfig(
     format="%(asctime)s - %(levelname)s - %(message)s", level=logging.INFO)
 
+# Poll interval to update workers on chain with those in kv store
+WORKER_REFRESH_INTERVAL = 30
+
 
 class BaseConnector(BlockchainConnectorInterface):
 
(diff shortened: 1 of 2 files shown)


blockchain_connector/common/connector_common/base_connector.py
```diff
@@ -81,24 +84,32 @@ def sync_registries(self):
...
+        # That will be a 2-way sync. As of now, only one way.
+        while True:
+            try:
+                self._active_worker_ids = \
+                    self._worker_delegate.lookup_workers_in_kv_storage()
+                # Lookup worker ids from blockchain
+                chain_worker_ids = self._worker_delegate.\
+                    lookup_workers_onchain()
+                # Add/Update workers to chain
+                self._worker_delegate.add_update_worker_to_chain(
```
Ram-srini, line 103: This loop will to call worker_update every time for the same worker. https://github.com/hyperledger/avalon/blob/f4d9f6a3b984f3c1c4114afb73d4c29068af1ef8/blockchain_connector/common/connector_common/worker_delegate.py#L104

rranjan3: Yes @Ram-srini we had a discussion over this. We need to build the intelligence to not update if things haven't changed. A simple way would be to hold a map of workers in-memory and check against it. We could calculate a hash of the JSON string (representing a worker from KV) to avoid field by field matching. 
Then at every poll, we can simply fetch worker from kv & match hashes.
```
dict = {} # (worker_id->hash)
for( w_id in kv):
    w_details_json = kv.retrieve(wo_id)
    w_hash = _get_hash(json.dumps(w_details_json))
    if w_id not in dict:
        w_registry.add(w_id, w_details_json)
    else:
        if w_hash != dict[w_id]:
            w_registry.update(w_id, w_details_json)
            w_registry.set_status(w_id, w_details_json["status"])
```

Ram-srini: Do you want to implement it in a separate PR?

rranjan3: No @Ram-srini . Will update this PR itself.

blockchain_connector/common/connector_common/worker_delegate.py
```diff
@@ -101,44 +104,58 @@ def add_update_worker_to_chain(self, wids_in_kv, wids_onchain):
...
-                            to blockchain".format(wid))
-                else:
-                    logging.info("Updated worker {} to blockchain".format(wid))
+                if wid not in self._worker_details_map or \
+                        self._worker_details_map[wid] != details:
+                    result = self._worker_instance.worker_update(
+                        worker_id, details)
+                    if result != ContractResponse.SUCCESS:
+                        logging.error("Error while updating worker {}"
+                                      " to blockchain".format(wid))
```
manju956, line 113: You could return if contract response is not success. Otherwise we will end up updating worker status to ACTIVE.

PR #626, opened 2020-07-14
convert methods.json file to python file

Ram-srini:
so that it can be part of python package


sdk/avalon_sdk/connector/blockchains/fabric/chaincode_methods.py
```diff
@@ -1,23 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Dictionary containing all fabric chain code names
+and their API names it supports along with their arguments
+"""
+cc_methods = {
```
manojgop, line 19: My suggestion is to to keep this as json file since that's more natural for the contents of this file. To make json file as part of python package, we can use the `data_files` parameter in `setup()` in `setup.py`

manojgop: `package_data`  in setup() is another option to consider instead of `data_files`

Ram-srini: I know about the data_files in setup.py, But here chain code methods are tightly coupled with code, it is not configuration file and it is not changing frequently.

manojgop: I see. data files need be configuration files. You could put an y files that you wamt to make part of the python package.  Since the PR comment mentioned this change is done "so that it can be part of python package", my suggestion was to achieve that we need not convert json file to python file.

manojgop: Another cons I see with is this introduces a global variable string which is pretty heavy weight. This could be avoided by loading the contents of json file when required.

Ram-srini: Content of json is used to validate the chain code APIs. It shouldn't be kept in the external json file. It should in align with code. It is more efficient than file read, better have to it in a dictionary to validate the calls.

manojgop: Irrespective of where it's kept, python file or json the chain code ABI has to align with the code which uses it. If reading is inefficient (let's say due to file size) we can read it once and store. I assume if this file is kept as part of package we can read it from local file system (and no network I/O is involved).  

Is there any possibility you can avoid the global variable ?

Ram-srini: I can change it const variable(with all CAPS)

manojgop: That would still make it global right :-) . If there is no better approach let's keep it current way though the string looks big and consume lot of memory. I like file read approach to load it when required and avoid global variable. But I guess you don't prefer it :-).

Ram-srini: Created a class for it.

PR #629, opened 2020-07-15
Optimize Python worker Graphene Manifest.

manojgop:
- Move manifest files to manifest directory and update build script.
- Reduce enclave size to improve startup time of python worker
  in Graphene-SGX.
- Allow only tmp directory in gcc,ld and collect2 manifest files.
  These are child manifest files used by pycryptodomex and
  these programs uses tmp directory.


examples/graphene_apps/python_worker/graphene/manifest/python.manifest
```diff
@@ -1,5 +1,9 @@
 sgx.allow_file_creation = 1
-sgx.enclave_size = 1G
+# Set the virtual memory size of the SGX enclave. For SGX v1, the enclave
+# size must be specified during upfront. If Python worker needs more
```
rranjan3, line 3: specified during upfront -> specified upfront

PR #634, opened 2020-07-16
Abstract docker compose files

rranjan3:


rranjan3:
> Since you moved the docker-compose files to docker/compose folder, I see you haven't changed the context in those files accordingly. The Dockerfile path in docker-compose file is relative to compose file location.

True @pankajgoyal2 partly. The context is taken care because the internal docker files cannot be run without the default. Not the conotext from default is replicated in all other services when files are chained and run. Which means the context still remains same. Only when a user explicitly goes out of way to build an images from elsewhere should he face challenges that too only when he decides not to use  the default files. With the defaults, he should not be able to build images because of dependencies on services from default docker-compose.yaml.

examples/graphene_apps/python_worker/README.md
```diff
@@ -107,7 +107,7 @@ https://creativecommons.org/licenses/by/4.0/
 
 - To run python worker as a docker container in Graphene-SGX environment and to use a test application to send work order requests, execute the following command from [python_worker](http://github.com/hyperledger/avalon/tree/master/examples/graphene_apps/python_worker) directory.
 
-  `docker-compose -f docker-compose.yaml -f docker-compose-graphene-sgx.yaml up`
+  `docker-compose -f docker-compose.yaml -f docker/compose/avalon-graphene-sgx.yaml up`
```
manojgop, line 110: Please don't change this. This is testing with out Avalon using test client.

pankajgoyal2:
> > Since you moved the docker-compose files to docker/compose folder, I see you haven't changed the context in those files accordingly. The Dockerfile path in docker-compose file is relative to compose file location.
> 
> True @pankajgoyal2 partly. The context is taken care because the internal docker files cannot be run without the default. Not the conotext from default is replicated in all other services when files are chained and run. Which means the context still remains same. Only when a user explicitly goes out of way to build an images from elsewhere should he face challenges that too only when he decides not to use the default files. With the defaults, he should not be able to build images because of dependencies on services from default docker-compose.yaml.

I see. So, the context is set based on the location of base compose file (first in sequence of docker-compose command).

examples/graphene_apps/python_worker/README.md
rranjan3: Updated.

PR #643, opened 2020-07-27
Cleaning Dockerfile from openssl dependency

pankajgoyal2:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -65,68 +65,6 @@ RUN apt-get update \
 RUN pip3 install setuptools
 
 
-# -------------=== Build openssl_image ===-------------
-
-#Build openssl intermediate docker image
-FROM ubuntu:bionic as openssl_image
-
-RUN apt-get update \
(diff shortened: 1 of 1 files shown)


rranjan3 (approved):

docker/Dockerfile
```diff
@@ -65,68 +65,6 @@ RUN apt-get update \
 RUN pip3 install setuptools
 
 
-# -------------=== Build openssl_image ===-------------
```
manojgop, line 68: Does verify report in Avalon shell use openSSL. @manju956 @pankajgoyal2  can you please confirm if verify report will work if we remove openSSL (untrusted) from Avalon shell docker image

Ram-srini: Generic client uses it to verify the AVR report of worker.

manju956: Verify_report module is built as separate python package. It depends on openssl as client invokes untrusted cpp code to do verification of attestation data.

danintel (approved):

PR #658, opened 2020-08-13
Changes to verify MR enclave value in client side

Ram-srini:
1. verify IAS report utility which verifies MR enclave
2. generic client shows the usage.


common/cpp/verify_ias_report/ias_attestation_util.cpp
```diff
@@ -54,3 +57,24 @@ bool verify_quote(const std::string& ias_report, int group_out_of_date_is_ok) {
     return quote_status;
 }
 
+bool verify_mr_enclave_value(const std::string& enclave_quote_body,
```
manju956, line 60: Please add function comments

common/verify_report_utils/verify_report/verify_attestation_report.py
```diff
@@ -47,4 +47,17 @@ def verify_attestation_report(enclave_info):
...
     logger.info("Enclave IAS report signature verification passed")
+
+    if mr_enclave != "":
+        v_report = json.loads(
+            enclave_info["proof_data"]["verification_report"])
+        mr_check = verify_report_util.verify_mr_enclave_value(
+            v_report["isvEnclaveQuoteBody"],
+            mr_enclave)
+        if mr_check is False:
+            logger.info("MR enclave value check failed")
```
manju956, line 58: Please change log level to error.

docker/Dockerfile
```diff
@@ -102,7 +102,9 @@ FROM ubuntu:bionic as common_cpp_image
 RUN apt-get update \
  && apt-get install -y -q \
     pkg-config \
+    wget \
     cmake \
+    libprotobuf-dev \
```
manju956, line 107: We dont need protobuf. please remove

manju956 (changes requested):

danintel (approved):

docker/Dockerfile
Ram-srini: It is required otherwise below dpkg complain while installing packages.

manju956: Ok

common/cpp/verify_ias_report/ias_attestation_util.cpp
Ram-srini: Done

manju956 (approved):

common/cpp/verify_ias_report/CMakeLists.txt
```diff
@@ -42,19 +42,22 @@ add_custom_command(OUTPUT  ${PROJECT_GENERATED_IAS_SOURCES}
 # Untrusted Verify-Ias-Report Library
 ################################################################################
 
+SET(SGX_SDK "$ENV{SGX_SDK}")
+SET(SGX_SDK_INCLUDE ${SGX_SDK}/include)
```
manojgop, line 46: Do we need SGX SDK dependency for untrusted part.? SGX SDK dependency previously was based on the condition `if(NOT UNTRUSTED_ONLY)`.

common/cpp/verify_ias_report/ias_attestation_util.cpp
```diff
@@ -54,3 +64,30 @@ bool verify_quote(const std::string& ias_report, int group_out_of_date_is_ok) {
...
+     **/
+    if (mr_enclave.size() != 0) {
+        /* Extract ReportData and MR_ENCLAVE from isvEnclaveQuoteBody
+           present in Verification Report */
+	ByteArray quote_bytes = Base64EncodedStringToByteArray(
+	    enclave_quote_body.c_str());
+        sgx_quote_t* quote_body = reinterpret_cast<sgx_quote_t*>(
+            quote_bytes.data());
+        sgx_report_body_t* report_body = &quote_body->report_body;
+        sgx_measurement_t mr_enclave_from_report = *(&report_body->mr_enclave);
```
manojgop, line 83: The issue with above approach is it brings SGX SDK dependency on common/cpp module. So we need to install SGX SDK in client Dockerfile ? ALl this is done just to extract the mrencalve value from quote.

manojgop (changes requested):

common/cpp/verify_ias_report/ias_attestation_util.cpp
manojgop: I would suggest to do mrenclave verification at python layer either using 
1) Python struct - similar to 
    https://github.com/hyperledger/sawtooth-poet/blob/master/common/sawtooth_poet_common/sgx_structs/_sgx_quote.py
https://github.com/hyperledger/sawtooth-poet/blob/master/common/sawtooth_poet_common/sgx_structs/_sgx_report_body.py
https://github.com/hyperledger/sawtooth-poet/blob/master/common/tests/test_sgx_structs/tests.py

or

2) Extract the mrenclave value as bytes from a fixed offset in the IAS report

manojgop (changes requested):

common/cpp/verify_ias_report/ias_attestation_util.cpp
Ram-srini: Since we are using python client for doing to MR enclave check and we are depend on C++ sgx sdk. Is there problem if we bring SGX sdk dependency in common/cpp? Anyway we are using sgx sdk for verifying signature and quote verification. Porting C++ structure to python is not right thing to do. It may lead to inconsistencies. I am not clear on option 2.

manojgop: One of design goals of Avalon is to hide SGX complexities from Avalon clients. If we make Avalon clients install SGX SDK we are moving away from that goal.  Sawtooth PoET for example solves this by doing AVR verification at Python layer and it uses the python struct. Another option is used a fixed offset in AVR quote (based on IAS attestation API spec)
Also bringing in SGX SDK dependency on Avalon client just to do MREnclave verification is not necessary.

common/cpp/verify_ias_report/ias_attestation_util.cpp
```diff
@@ -13,20 +13,30 @@
...
 bool verify_ias_report_signature(const std::string& signing_cert_pem,
-                                   const std::string& ias_report,
-				   const std::string& ias_signature) {
+                                 const std::string& ias_report,
+                                 const std::string& ias_signature) {
+    /* Verify IAS report signature
+     * @param signing_cert_pem signing certificate
+     * @param ias_report attestion report
+     * @param ias_signature attestation report signature
+     * Returns true if signature verification success
```
danintel, line 37: Change Return to @return

common/cpp/verify_ias_report/ias_attestation_util.cpp
```diff
@@ -54,3 +64,30 @@ bool verify_quote(const std::string& ias_report, int group_out_of_date_is_ok) {
     return quote_status;
 }
 
+bool verify_mr_enclave_value(const std::string& enclave_quote_body,
+		             const std::string& mr_enclave) {
+    /* Verify MR enclave in the attestation
+     * report and compare with the value passed
+     * @param enclave_quote_body Enclave quote body
+     * @param mr_enclave MR enclave value in hex format
+     * Return true if comparision matches otherwise false
```
danintel, line 73: Change Return to @return

danintel (approved):
Use @return in your block comments

PR #662, opened 2020-08-17
Fix to fabric proxy model test failure.

Ram-srini:
Removed try catch for the run_until_complete call
Replaced asyncio.wait with asyncio.gather

### DIFF
--- a/blockchain_connector/common/connector_common/base_connector.py
+++ b/blockchain_connector/common/connector_common/base_connector.py
@@ -157,14 +157,11 @@ class BaseConnector(BlockchainConnectorInterface):
         """
         event_loop = asyncio.get_event_loop()
         logging.info("Blockchain Connector service started")
-        try:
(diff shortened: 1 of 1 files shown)


blockchain_connector/common/connector_common/base_connector.py
```diff
@@ -157,14 +157,11 @@ def start(self):
...
-            event_loop.close()
+        event_loop.run_until_complete(self.sync())
 
     async def sync(self):
         """
         Wrapper function to wait for the coroutines to finish. It simply
         waits for the sync_workers() & sync_work_orders() to complete.
         """
-        await asyncio.wait([self.sync_workers(), self.sync_work_orders()])
+        await asyncio.gather(self.sync_workers(), self.sync_work_orders())
```
rranjan3, line 167: Should not be an issue ideally unless we group the co-routines or need to handle the futures returned by the co-routines. But, let it be for now.

PR #676, opened 2020-09-10
Update wget to curl to download IAS certs

rranjan3:

### DIFF
--- a/common/cpp/verify_ias_report/build_ias_certificates_cpp.sh
+++ b/common/cpp/verify_ias_report/build_ias_certificates_cpp.sh
@@ -35,7 +35,7 @@ Cleanup () {
 trap 'echo "**ERROR - line $LINENO**"; Cleanup; exit 1' HUP INT QUIT PIPE TERM ERR
 
 #get certificate from Intel
-wget https://software.intel.com/sites/default/files/managed/7b/de/RK_PUB.zip
+curl https://software.intel.com/sites/default/files/managed/7b/de/RK_PUB.zip -o RK_PUB.zip
 test -e RK_PUB.zip
 echo "Zipped certificated downloaded"
 


pankajgoyal2 (approved):

common/cpp/verify_ias_report/build_ias_certificates_cpp.sh
```diff
@@ -35,7 +35,7 @@ Cleanup () {
 trap 'echo "**ERROR - line $LINENO**"; Cleanup; exit 1' HUP INT QUIT PIPE TERM ERR
 
 #get certificate from Intel
-wget https://software.intel.com/sites/default/files/managed/7b/de/RK_PUB.zip
+curl https://software.intel.com/sites/default/files/managed/7b/de/RK_PUB.zip -o RK_PUB.zip
```
manju956, line 38: Why is wget replaced by curl? Does this address any issue?

manju956: I just noticed, wget is failing to download certificates from the url and with this change, its downloading certificate.

manju956 (approved):

Ram-srini (approved):

common/cpp/verify_ias_report/build_ias_certificates_cpp.sh
rranjan3: Yes @manju956 . Though I am not too sure what has caused this to happen suddenly. It could be that the remote server is checking for the User-Agent explicitly and allowing only curl.

danintel: That is strange. There is also a `wget -U agent-string` option to fake the User-Agent: header line.

PR #682, opened 2020-09-27
Reduce enclave size in graphene python worker manifest files

manju956:
- sgx enclave size in graphene python worker is reduced from 512M to 256M as higher enclave size results in 
  Out-Of-Memory issue leading to crash python worker container and crashing docker daemon.


danintel (approved):
Does this enclave size have anything to do with real Intel SGX hardware enclave sizes?
Because the maximum is 128 Mbytes on Intel Xeon E3, for example.

Ram-srini (approved):

rranjan3 (approved):

Issue #693, opened 2020-10-12
Enclave manager crashes when work order processing is not completed and work order receipts is turned on.


sourabh-natesh:
When the work order has not been completed, the enclave manager crashes with the following message: 

> avalon-wpe                | [09:28:51 ERROR   avalon_enclave_manager.work_order_processor_manager] Error while processing work-order; shutting down enclave manager
avalon-wpe                | [09:28:51 ERROR   avalon_enclave_manager.work_order_processor_manager] Exception: <class 'TypeError'> args ('must be str, not dict',) details must be str, not dict

For example, this happens when I pass a null pointer to the ThrowIfNull function in the enclave manager code. 
The error is happening due to an improper concatenation of string and dictionary while updating the work order receipt. The error is in line _163_ of the file _work_order_receipt.py._

PR #694, opened 2020-10-12
Fixing issue with updating work order receipt

sourabh-natesh:
Pushing the fix for the issue #693 .

In the file work_order_receipt.py, in line 163, while creating the update receipt - we concatenate the update data to the work order receipt string. The issue is that if the receipt create status has the value FAILED, the update data will be a dictionary, and hence, concatenating it to the string will cause the enclave manager to crash. 

Have pushed code to convert this update data into a string so that it can be concatenated. 

Signed-off-by: Sourabh Natesh sourabh.natesh@intel.com

### DIFF
--- a/sdk/avalon_sdk/work_order_receipt/work_order_receipt.py
+++ b/sdk/avalon_sdk/work_order_receipt/work_order_receipt.py
@@ -135,7 +135,7 @@ class WorkOrderReceiptRequest():
         Returns:
         JSON RPC work order update receipt request of type dictionary
         """
-        data = update_data
+        data = json.dumps(update_data)
         if update_type in [ReceiptCreateStatus.PROCESSED.value,
                            ReceiptCreateStatus.COMPLETED.value]:
             # Work Order Receipt status is set to be completed or processed,


danintel:
You have no commit message, so the build is also failing. Put the message above, including your DCO signature line, in the commit.  You can amend your commit with
`git commit --amend`
Then force push the commit to your github workspace:
`git push -f <your workspace> <github workspace>`

rranjan3 (approved):
Looks good

sdk/avalon_sdk/work_order_receipt/work_order_receipt.py
```diff
@@ -135,7 +135,7 @@ def update_receipt(self, work_order_id, update_type,
         Returns:
         JSON RPC work order update receipt request of type dictionary
         """
-        data = update_data
+        data = json.dumps(update_data)
```
Ram-srini, line 138: At line 143 also we are converting to serialized string, which is duplicate. instead of line 138 we can use wo_resp_str at 143 and remove serialization at 138.

PR #707, opened 2020-11-04
Fix 'RuntimeError: This event loop is already running' bug of asyncio in ethereum-ganache model

xliamren:
fix 'RuntimeError: This event loop is already running' bug of asyncio in ethereum-ganache model

Signed-off-by: PlyTools <qianren1024@gmail.com>


rranjan3:
Trying to fix the same issue in #703 .  Do you see any issues in the other approach? If no, we could make a similar change in Fabric connector as well. Removing the dependency on nest_asyncio.

PR #724, opened 2020-11-23
Pass docker build args for OS and distro

manju956:

### DIFF
--- a/docker/compose/avalon-multi-pool.yaml
+++ b/docker/compose/avalon-multi-pool.yaml
@@ -36,6 +36,8 @@ services:
         - https_proxy
         - no_proxy
         - ENCLAVE_TYPE=kme
+        - IMAGE=$IMAGE
+        - DISTRO=$DISTRO
     environment:
       - http_proxy
       - https_proxy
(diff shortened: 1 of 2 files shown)


docker/compose/avalon-pool.yaml
```diff
@@ -74,6 +76,8 @@ services:
         - https_proxy
         - no_proxy
         - ENCLAVE_TYPE=wpe
+        - IMAGE=$IMAGE
```
rranjan3, line 79: Would not be required. Will be taken up from default docker-compose.yaml.

docker/compose/avalon-multi-pool.yaml
```diff
@@ -79,6 +81,8 @@ services:
         - https_proxy
         - no_proxy
         - ENCLAVE_TYPE=wpe
+        - IMAGE=$IMAGE
```
rranjan3, line 84: Would rely on defaults from docker-compose.yaml. Double check for all other attributes as well.

rranjan3 (commented):
Minor changes required.

docker/compose/avalon-pool.yaml
manju956: Yes. Removed it along with other duplicate attriutes.

docker/compose/avalon-multi-pool.yaml
manju956: Removed all the duplicate attributes for the overriden service.

rranjan3 (approved):

pankajgoyal2 (approved):

PR #731, opened 2020-12-12
 Implemented Key Sharing and Seperation Config id Feature for Singleton.

karthikamurthy:
Signed-off-by: Karthika Murthy <karthika.murthy@intel.com>

### DIFF
--- a/tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/enclave.cpp
+++ b/tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/enclave.cpp
@@ -70,7 +70,8 @@ namespace tcf {
         // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
         void Enclave::Load(
             const std::string& inEnclaveFilePath,
-            const Base64EncodedString& inSealedEnclaveData) {
+            const Base64EncodedString& inSealedEnclaveData,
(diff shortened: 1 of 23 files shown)


tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/enclave.cpp
```diff
@@ -147,7 +157,9 @@ namespace tcf {
 
                 // First attempt to load the enclave executable
                 sgx_status_t ret = SGX_SUCCESS;
-                ret = tcf::sgx_util::CallSgx([this, flags, &token] () {
+                if(this->_kss_config[0] == NULL){
```
Ram-srini, line 160: why are we checking only index 0, is that sufficient?

tc/sgx/trusted_worker_manager/enclave_untrusted/enclave_bridge/enclave.cpp
```diff
@@ -160,7 +172,30 @@ namespace tcf {
                     10,  // retries
                     250  // retryWaitMs
                     );
-                tcf::error::ThrowSgxError(ret, "Unable to create enclave.");
+                    tcf::error::ThrowSgxError(ret, "Unable to create enclave.");
+                    
+                } else {
+                    tcf::Log(TCF_LOG_INFO, "Enclave::sgx_create_enclave_ex called" );
+                    void *enclave_ex_p[32] = { 0 };
```
Ram-srini, line 179: use constant instead of hard coded value 32, may be sgx sdk or create one.

